PR #2, opened 2018-01-25
Add Sphinx documentation site

kylef:
This PR adds the base Sphinx documentation with installation and basic usage. This is an MVP. I have full API reference and documentation to publish, I will incrementally PR the separate changes.

messages image 1270598827


Almad:
@kylef I think we should go with consistency. Open to which theme we should use—probably something for brainstorm and/or team lead sync.

Also //cc @michalholasek @w-vi @abtris

PR #4, opened 2018-03-30
Add API Reference documentation

kylef:
I changed the branch that readthedocs is building so we can test this branch (will restore after review).

This branch is built and published to http://api-elements-js.readthedocs.io/en/latest/elements.html. This includes decent documentation coverage for core minim elements. The API Description and Parse Result elements are lacking somewhat, but that is separate topic than this PR, this PR sets up the API Reference. Once we add further documentation to minim-api-description it will be shown here.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,2 @@
+build:
+    image: latest
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "author": "Apiary",
   "license": "MIT",
   "dependencies": {
-    "minim": "^0.20.4",
+    "minim": "git+https://github.com/refractproject/minim.git",
     "minim-parse-result": "^0.10.0"
   },
   "devDependencies": {
--- a/docs/conf.json
+++ b/docs/conf.json
@@ -1,0 +1,10 @@
+{
+  "source": {
+    "include": [
+      "../node_modules/minim/lib/primitives",
+      "../node_modules/minim/lib/elements",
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
   "author": "Apiary",
   "license": "MIT",
   "dependencies": {
-    "minim": "^0.20.4",
+    "minim": "git+https://github.com/refractproject/minim.git",
```
kylef: JSDoc actually crashes due to not handling the parameters which I fixed in https://github.com/refractproject/minim/pull/171. Have to use this version of minim so docs build until minim master is released. I'd continue for now with it as git dependency and release minim once we're ready (I likely want to change further documentation in separate PRs).

.readthedocs.yml
```diff
@@ -0,0 +1,2 @@
+build:
+    image: latest
```
kylef: @hj @michalholasek Here's a trick to use readthedocs image which contains jsdoc and node, npm. I noticed that Dredd is installing node itself, you can simplify Dredd readthedocs configuration by using this method instead.

NOTE, as I am documenting my dependencies, all I needed to do is `npm install` inside docs/conf.py

package.json
pksunkara: We should probably release minim master instead of doing this.

.readthedocs.yml
honzajavorek: This is awesome!

PR #56, opened 2019-01-09
Test API Blueprint parser via Fury interface

kylef:
Instead of testing the interface directly, because we should be testing with Minim namespace and version version from Fury instead of creating our own which may differ. This helps us catch bugs. In the future we will no longer depend on Minim so this is refactoring to allow that.

These are half the changes from https://github.com/apiaryio/fury-adapter-apib-parser/pull/30, I wanted to make this refactoring land in master so the amount of code we need to change to make a Drafter 4 update is very minimal.

### DIFF
--- a/packages/fury-adapter-apib-parser/test/validate-test.js
+++ b/packages/fury-adapter-apib-parser/test/validate-test.js
@@ -1,14 +1,15 @@
 const { expect } = require('chai');
-const { Namespace } = require('minim');
-const { validate } = require('../lib/adapter');
+const { Fury } = require('fury');
+const adapter = require('../lib/adapter');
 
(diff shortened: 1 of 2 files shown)


packages/fury-adapter-apib-parser/test/validate-test.js
```diff
@@ -53,7 +54,7 @@ describe('API Blueprint validation', () => {
                         content: [
                           {
                             element: 'number',
-                            content: 0,
+                            content: 11,
```
kylef: NOTE, Added FORMAT: 1A so we're detecting document as API Blueprint and thus that is 11 additional characters at the start.

PR #58, opened 2019-01-09
Use Drafter 4

kylef:
Superceeds https://github.com/apiaryio/fury-adapter-apib-parser/pull/30

### DIFF
--- a/packages/fury-adapter-apib-parser/CHANGELOG.md
+++ b/packages/fury-adapter-apib-parser/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Fury API Blueprint Parser Changelog
 
+## Master
+
+This update now uses drafter-npm 2.0.0-pre.1. Please see [drafter-npm
+2.0.0-pre.1](https://github.com/apiaryio/drafter/releases/tag/v2.0.0-pre.1) for
+the list of changes.
+
 ## 0.12.0 (2018-12-21)
 
 ### Breaking
--- a/packages/fury-adapter-apib-parser/CHANGELOG.md
+++ b/packages/fury-adapter-apib-parser/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Fury API Blueprint Parser Changelog
 
-## Master
+## 0.13.0-beta (2019-01-10)
 
 This update now uses drafter-npm 2.0.0-pre.1. Please see [drafter-npm
 2.0.0-pre.1](https://github.com/apiaryio/drafter/releases/tag/v2.0.0-pre.1) for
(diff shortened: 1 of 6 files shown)


kylef:
I am hesistant for us to merge this, because we may still need to make other releases of the Fury toolchain especially for OAS3 support and I am not sure how long rolling out Drafter 4 to other places will take. We may need to make future releases of the adapter without Drafter 4 which this PR would prevent.

Although, I think it would be great to be able to test this in Dredd, Apiary UI and be able to test it out in Apiary which this PR would allow. Therefore I was thinking about making this beta release from the unmerged branch. This means we can still release 0.13.0 and other versions without Drafter 4 and we have a pre-release to start testing and if it goes well we and merge this, and make a final 0.13.0 release with the Drafter 4 pre-releases.

What do you think?

honzajavorek:
Sounds like juggling with two big changes 🙈

If you think the API Blueprint adapter is going to need changes related to introducing OAS3 support (why? is versioning of the whole Fury toolchain now synchronized?), then I guess what you wrote makes sense.

Otherwise I can imagine OAS3 happening in the OAS3 adapter while APIB improvements happen in the API adapter. The only changes spanning both would be changes to Fury itself (can those be postponed when we're sure these two things work well?) or changes to API Elements (but I guess the spec and minim should be somewhat stable now?).

kylef:
If we need to make changes in Minim, then it requires us to release a new version of the entire Fury toolchain at the moment since we're at a `beta` release of it and it's API is coupled to Minim which is < 1.0 and thus semver dictates that almost all changes except small bug fixes will require us to bump it in Fury, and release a new version of fury and all adapters. One example is Minim 0.22.0 (https://github.com/refractproject/minim/blob/1b9d9ee66689839cd9a5246741e0d10a7a945402/CHANGELOG.md#0220) which is an enhancement (and potentially breaking for some consumers of Fury that use Ramda) which required us to make an entire release of Fury and all the adapters for semver reasons, and rightly so.

> If you think the API Blueprint adapter is going to need changes related to introducing OAS3 support (why? is versioning of the whole Fury toolchain now synchronized?), then I guess what you wrote makes sense.

The API Blueprint adapter won't need changes related to OAS3. However, if we need to release new version of Minim for any reason with potentially breaking changes, that means a new Fury toolchain. If we had to update Minim for OAS 3 reasons (I don't think we will but it could happen) then we need Fury release for OAS3 completion. In that case we'd have to release all fury adapters including fury-adapter-apib-parser in the process due to semver. If we're half way through rolling out fury-adapter-apib-parser with Drafter 4 it may prevent us rolling out other changes like OAS3.

honzajavorek:
I didn't realize the minim/semver situation. Drafter 4 pre-releases make sense then, I guess.

kylef:
I discovered that there was a problem in the last Protagonist release where some  files are missing in the NPM release, we will need to make another release of Protagonist and Drafter-NPM and update it here.

Issue #20, opened 2018-11-28
Coverage for all packages


pksunkara:

kylef:
There is coverage in coveralls.

PR #60, opened 2019-01-17
Use Circle CI instead of Travis CI

kylef:


.travis.yml
```diff
@@ -1,23 +0,0 @@
...
-script:
-  - npm run lint
-  - npm test
-before_script:
-  - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
-  - chmod +x ./cc-test-reporter
-  - ./cc-test-reporter before-build
-after_script:
-  - npm run cover
-  - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
```
kylef: This is actually broken and it is silently outputting an error. There is https://github.com/apiaryio/api-elements.js/issues/20 to track adding coverage.

.circleci/config.yml
```diff
@@ -0,0 +1,44 @@
...
+    <<: *defaults
+
+  test-node10:
+    <<: *defaults
+
+  lint:
+    <<: *defaults
+    steps:
+      - checkout
+      - <<: *install
```
honzajavorek: oh this is a cool trick

PR #88, opened 2019-01-18
Support schema referencing in media type object

kylef:


packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
```diff
@@ -31,9 +31,17 @@ function parseComponentsObject(context, element) {
   const validateIsObject = key => R.unless(isObject,
     createWarning(namespace, `'${name}' '${key}' is not an object`));
 
+  const parseSchemaMember = member => pipeParseResult(namespace,
+    R.compose(parseSchemaObject(context), getValue),
+    (dataStructure) => {
+      // eslint-disable-next-line no-param-reassign
```
kylef: I understand what this eslint rule is meant to protect against and in a lot of cases I agree with it. However, in this particular line I think this makes sense, we're modifying the dataStructure to attach the ID.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
```diff
@@ -31,9 +31,17 @@ function parseComponentsObject(context, element) {
...
+      dataStructure.content.id = member.key.clone();
```
kylef: In https://github.com/apiaryio/api-elements.js/pull/86 I moved this logic a bit too far away and into parseOpenAPI and I think instead it should be here inside the components parser so the components contain the ID used for referencing.

packages/fury-adapter-oas3-parser/lib/parser/parseReference.js
```diff
@@ -8,7 +8,19 @@ function isReferenceObject(element) {
 
 function parseReference(component, parser, context, element) {
   if (isReferenceObject(element)) {
-    return parseReferenceObject(context, component, element);
+    const result = parseReferenceObject(context, component, element);
```
honzajavorek: Took me a moment to realize that this isn't a smelly variable name, but a variable for "parse result". Perhaps `parseResult` would cognitively lead better to realizing what's in it?

kylef: I'll prepare a separate PR changing the variable name, this is a frequent pattern in most files. I'll sed the code base and do it all.

PR #90, opened 2019-01-18
Move validating object into parseObject

kylef:


packages/fury-adapter-oas3-parser/lib/parser/annotations.js
```diff
@@ -38,6 +38,10 @@ function createMemberValueNotBooleanError(namespace, path, member) {
 }
 
 function validateObjectContainsRequiredKeys(namespace, path, requiredKeys, object) {
+  if (!isObject(object)) {
```
kylef: Since `validateObjectContainsRequiredKeys` can sometimes be ran before the `isObject` validation it would fail we need protection against it.

This is an interim patch because I was planning to add `requiredKeys` behaviour to `parseObject` itself.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
```diff
@@ -41,7 +41,7 @@ function parseComponentsObject(context, element) {
 
   const parseSchemasObject = pipeParseResult(namespace,
     validateIsObject('schemas'),
-    parseObject(context, parseSchemaMember));
+    parseObject(context, name, parseSchemaMember));
```
kylef: There is still a `validateIsObject` check above because it has a different warning format in regards to the name. I'll adapt this separately, perhaps something like making it so the name may be an array of names:

```js
parseObject(context, [name, 'schemas'], parseSchemaMember)
```

honzajavorek: I'd use composition to create something like `parsePropertyAsObject` where the validation of the property with the property name would happen. Just quick a thought from the top of my head.

kylef: Good idea!

PR #91, opened 2019-01-18
feat(oas3): support description in response object

### LINKED ISSUE #62: Support 'description' in response object parser
`parseResponseObject` should handle `description` and place it into the resultant HttpResponse element.

```
lib/parser/oas/parseResponseObject.js
```

klokane:
resolve #62

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseResponseObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseResponseObject.js
@@ -11,9 +11,11 @@ const {
 const parseObject = require('../parseObject');
 const parseMediaTypeObject = require('./parseMediaTypeObject');
 
+const parseCopy = require('../parseCopy');
+
 const name = 'Response Object';
 const unsupportedKeys = [
-  'description', 'headers', 'links',
+  'headers', 'links',
 ];
 const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
 
@@ -58,6 +60,7 @@ function parseResponseObject(context, element) {
 
   const parseMember = R.cond([
     [hasKey('content'), R.compose(parseContent, getValue)],
+    [hasKey('description'), parseCopy(context, name, false)],
 
     [isUnsupportedKey, createUnsupportedMemberWarning(namespace, name)],
 
@@ -71,21 +74,21 @@ function parseResponseObject(context, element) {
   const parseResponse = pipeParseResult(namespace,
     parseObject(context, name, parseMember),
     (responseObject) => {
-      const responses = responseObject.get('content');
+      // Try to fecth responses from the media type parsing
+      // if not, create empty HttpResponse
+      const responses = R.or(responseObject.get('content'), [new namespace.elements.HttpResponse()]);
 
-      if (responses) {
-        // If we have responses from the media type parsing, return those
-        // after attaching Response Object information
-        return new namespace.elements.ParseResult(responses.map((response) => {
-          response.statusCode = element.key.toValue();
-          return response;
-        }));
-      }
+      const description = responseObject.get('description');
 
-      // No media types defined in content, return empty response
-      const response = new namespace.elements.HttpResponse();
-      response.statusCode = element.key.toValue();
-      return response;
+      return new namespace.elements.ParseResult(responses.map((response) => {
+        response.statusCode = element.key.toValue();
+
+        if (description) {
+          response.push(description);
+        }
+
+        return response;
+      }));
     });
 
   return parseResponse(element.value);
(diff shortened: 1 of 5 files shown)


packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseResponseObject-test.js
```diff
@@ -36,19 +36,32 @@ describe('Response Object', () => {
...
+        description: [],
+      });
+
+      const result = parse(context, response);
+
+      expect(result).to.contain.warning("'Response Object' 'description' is not a string");
     });
   });
 
+
```
kylef: NIT: Empty new line added.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseResponseObject-test.js
```diff
@@ -36,19 +36,32 @@ describe('Response Object', () => {
       expect(result.length).to.equal(1);
       expect(result).to.contain.warning("'Response Object' response status code ranges are unsupported");
     });
+  });
+
+  describe('description', () => {
```
kylef:
```suggestion
  describe('#description', () => {
```

When the describe name is a property name or a function name itself instead of a string or class name it's common to represent this with a `#`, for example `#description`. You can see this in some of the other test names. Would you be able to update for consistency?

```
  #parsePathItemObject
    ✓ parses a path into a resource
    ✓ parses a path methods into a resource and transactions
    ✓ provides a warning when the path item object is non-object
    warnings for keys
      ✓ warns for $ref
      ✓ warns for a servers
      ✓ does not provide warning for Info Object extensions
      ✓ provides warning for invalid keys
    #summary
      ✓ warns when summary is not a string
      ✓ exposes summary as the title of the resource
    #description
      ✓ exposes description as a copy element in the resource
      ✓ warns when description is not a string
    #parameters
      ✓ warns when parameters is not an array
      path parameters
        ✓ exposes parameter in hrefVariables
        ✓ errors when parameter is not found in path
        ✓ errors when path variable not defined in parameters
      query parameters
        ✓ exposes query parameter in href
        ✓ exposes multiple query parameter in href
        ✓ exposes query parameter in hrefVariables
```

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseResponseObject-test.js
```diff
@@ -36,19 +36,32 @@ describe('Response Object', () => {
...
+    it('exposes description of response', () => {
```
kylef: This test isn't testing that the description in found within the response and the implementation isn't exposing the copy element.

We will need to push the copy element into the HttpResponse's in `parseResponseObject.js`, on line `~80` and `~87` where the status code is added to the HttpResponse object we should insert the copy element into these elements (at index 0). There is similiar code for this in `lib/parser/oas/parseOperationObject.js`:

```js
      const description = operation.get('description');
      if (description) {
        transition.push(description);
      }
```

In our case this will end up in multiple responses and thus we should make a `clone` of the copy element (`httpResponse.push(description.clone);`). That because it isn't good practise to put the exact same element instance multiple times in the tree as any consumers of the parser may do mutation which may incorrectly happen to other references to the same copy element.

The biggest reason though is that a consumer might use `.freeze()` on the element tree to make it immutable (and also add .parent propert accessors which falls down when the same element is reused because it can only have one parent element). (https://github.com/refractproject/minim/blob/1b9d9ee66689839cd9a5246741e0d10a7a945402/lib/primitives/element.js#L34)

Then we'll see the test fixtures include the copy elements (can be generated with `npm run test:integration:fix`).

kylef:
One other thing, the commit message uses the type `update`. Could you update this to be `feat` (https://www.conventionalcommits.org/en/v1.0.0-beta.2/) to follow the conventional commits we're /somewhat/ following. I've added some linting for the commit messages in https://github.com/apiaryio/api-elements.js/pull/92 so we can enforce these at CI level. I think it's also worth scoping these changes to `oas3` so it's easier to follow what the commits affect in the monorepo.

> feat(oas3): support 'description' in response object parser

klokane:
Updated version is ready for review.
This does not contain check if `description` for response is presented (according to OAS is `description` REQUIRED. 
Will be delivered in following PR

packages/fury-adapter-oas3-parser/lib/parser/oas/parseResponseObject.js
```diff
@@ -71,21 +74,21 @@ function parseResponseObject(context, element) {
...
-        }));
-      }
-
-      // No media types defined in content, return empty response
-      const response = new namespace.elements.HttpResponse();
-      response.statusCode = element.key.toValue();
-      return response;
+      // Try to fecth responses from the media type parsing
+      // if not, create empty HttpResponse
+      const responses = R.or(responseObject.get('content'), [new namespace.elements.HttpResponse()]);
```
kylef: Nice!

PR #92, opened 2019-01-18
Lint commit messages

kylef:
This PR adds commit message linting, this is similiar to the configuration in Dredd except as this is a monorepo I set it to require a scope of fixed value from one of the areas of the project.

- `project` - Refers to project in general, for example this PRs commits themselves.
- `cli` - for fury-cli
- `oas3` - OpenAPI 3 Parser
- `oas2` - OpenAPI 2 Parser 
- `apib` - API Blueprint Parser and Serialiser
- `apiaryb` - Apiary Blueprint Parser

We may need to give it a further think and perhaps add more permitted scopes in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,7 @@ jobs:
       - checkout
       - <<: *install
       - run: npm run lint
+      - run: npx commitlint --from origin/master
 
 workflows:
   version: 2
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,14 @@
+module.exports = {
+  extends: ['@commitlint/config-conventional'],
+  rules: {
+    'scope-enum': [2, 'always', [
+      'project',
+      'cli',
+      'oas3',
+      'oas2',
+      'apib',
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -33,6 +33,7 @@ jobs:
       - checkout
       - <<: *install
       - run: npm run lint
+      - run: npx commitlint --from origin/master
```
honzajavorek: Wouldn't it make sense to make this a part of `npm run lint`?

kylef: Unsure, it might have undesired effects (wrong branches etc as it's tied to git tree) and it's extremely slow:

```shell
$ time npx commitlint --from origin/master
...

        8.76 real         6.47 user         2.39 sys
```

That said, I personally don't use `npm run lint` command and instead only use lint plugins inside my editor. So I'll take whatever the consensus is for this command.

honzajavorek: I don't think there's a strong convention, it's up to every maintainer's preference.

PR #101, opened 2019-01-21
Support enum in Schema Object

kylef:


klokane:
@kylef how it is related to MSON enum? Does it should support same behavior or it "something completly different"

kylef:
@klokane OAS 3 "enum" is very similiar, the one difference is values are always fixed, hence I've added the "fixed" typeAttribute to every element type, except null as this value is inherantly fixed by default.

kylef:
We can merge, the changes are relatively small if anyone else has comments please do put them here and I can address separately.

PR #109, opened 2019-01-24
chore(oas3): Some cleanup regarding package structure

pksunkara:


honzajavorek:
The solution is to

```diff
- chore(oas3): Some cleanup regarding package structure 
+ chore(oas3): some cleanup regarding package structure 
```

pksunkara:
I have fixed it for now. But I want discussion on this. IMHO commit messages first line are supposed to be sentences and sentences start with a capital letter.

honzajavorek:
It's a convention specified by the conventional changelog: https://github.com/marionebl/commitlint/tree/master/@commitlint/config-conventional#subject-case

See the examples here https://www.conventionalcommits.org/en/v1.0.0-beta.2/#examples. Ultimately it comes from here: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#subject

We can discuss, but unless you make your own shareable commitlint config, there are no possible action points other than not to use conventional changelog.

PR #111, opened 2019-01-25
Respect generateSourceMap option

tjanc:
Filters out sourceMap attributes from non-annotation elements at the end of parsing.
Filtering at the end is suboptimal, but currently we are creating annotation source maps from their elements during parsing, so changing this might be a mess.

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
@@ -27,6 +27,45 @@ const unsupportedKeys = ['servers', 'security', 'tags', 'externalDocs'];
  */
 const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
 
+function filterSourceMaps(result) {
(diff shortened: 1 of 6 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
```diff
@@ -27,6 +27,45 @@ const unsupportedKeys = ['servers', 'security', 'tags', 'externalDocs'];
...
 
+function filterSourceMaps(result) {
+  if (isAnnotation(result)) {
+    return result;
+  }
+  if (result) {
+    if (!result.element) {
+      return result;
+    }
+    if (result.attributes) {
```
kylef: I'd recommend making this check `result._attributes` (same for meta `_meta`). That's because the underlying object elements returned by these properties are lazily created. If we are traversing an element with no attributes or meta it will cause the creation of two object elements which will have a performance impact.

Check out the lazy getters in https://github.com/refractproject/minim/blob/07c9435183081471bbc6a7a661b0ce0ca35d539e/src/primitives/element.js#L327

PR #112, opened 2019-01-25
Prefer example values when generating body in OAS 2

kylef:
In some cases, it seems like the logic to compute intermediary “JSON Schema” (from Swagger Schema) is failing to copy some of the definitions over thus causing a ref to fail and expand to `null` causing failure to render body.

I've had these changes stashed locally for a while as I've been unable to write a test that can actually test this behaviour. without disclosing private API Description Documents from a user's report.


kylef:
Sharing this here for now, I'll try figure out how to write some tests.

PR #113, opened 2019-01-27
Refactor parseObject to handle required keys validations

kylef:
This PR makes `parseObject` responsible for validating required keys, and it can do it both before and "after" each member has been parsed. This means if a required property (required member) cannot be parsed it can return a warning and `parseObject` can fail parsing the entire object.

Here's an example:

```js
const requiredKeys = ['in'];
const parseMember = R.cond([
  ['in', parseIn],
  ['required', parseRequired],
]);
const parser = parseObject(context, 'Parameter Object', parseMember, requiredKeys);
```

Now, if the require property `in` is missing, parseObject will create a warning upfront and prevent continuing parsing the object. If there is an `in` property, but it is invalid, then the "in" parser (`parseIn`) can fail (not return an element, returning a warning etc) then `parseObject` will fail parsing the object and return out. This makes it much simpler for us to create warnings that exit out certain object parsers for required properties.

This PR also makes use of this functionality to make the parameter 'in' property cause a warning for unsupported values.

I though about making some kind of composable "required" middleware function, but I cannot see any use of making it separate at the moment. We could change this in the future though.

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
@@ -67,7 +67,7 @@ function parseComponentsObject(context, element) {
     [R.T, createInvalidMemberWarning(namespace, name)],
   ]);
 
-  return parseObject(context, name, parseMember, element);
+  return parseObject(context, name, parseMember)(element);
 }
 
 
--- a/packages/fury-adapter-oas3-parser/CHANGELOG.md
+++ b/packages/fury-adapter-oas3-parser/CHANGELOG.md
@@ -12,6 +12,9 @@
   warning, previously using unsupported properties was emitting a warning that
   the properties were invalid.
 
+- The parser will no longer error out for unsupported parameter 'in' values,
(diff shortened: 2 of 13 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
```diff
@@ -67,7 +67,7 @@ function parseComponentsObject(context, element) {
     [R.T, createInvalidMemberWarning(namespace, name)],
   ]);
 
-  return parseObject(context, name, parseMember, element);
+  return parseObject(context, name, parseMember)(element);
```
kylef: Just to make it clearer for the reviewer, `parseObject` is a function that takes some arguments to construct an object parser, and it returns another function that is the final parser that accepts the single argument of element.

This was done to make it simpler to provide optional properties, this pattern happens elsewhere in Ramda and the code-base so it isn't specific to this case (see `R.pipe`, `R.compose` and `pipeParseResult`).

pksunkara:
Was thinking of this when reading the codebase.

PR #114, opened 2019-01-27
Support operation parameters

kylef:
This PR adds support for operation parameters. There is a missing part of finishing operation parameters off, thats regarding an error we emit due to the integration of path item and operation parameters. Validation of the path parameters only happens against parameters declare at the path item level. There will be a follow up PR which fixes additional cases but this is good enough for the purposes of iteration.


packages/fury-adapter-oas3-parser/test/integration/fixtures/petstore.json
```diff
@@ -1495,7 +1667,7 @@
           ]
         }
       },
-      "content": "'Operation Object' contains unsupported key 'parameters'"
+      "content": "'Parameter Object' contains invalid key 'schema'"
```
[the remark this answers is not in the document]
kylef: Resolved in https://github.com/apiaryio/api-elements.js/pull/116, will need to rebase either PR depending on what gets merged first.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
```diff
@@ -33,16 +33,40 @@ function createTransactions(namespace, member, operation) {
...
+
+  const query = parameters.get('query')
+    ? parameters.get('query')
+    : new namespace.elements.HrefVariables();
+
+  if (!path.isEmpty || !query.isEmpty) {
+    return path.concat(query);
+  }
+
+  return undefined;
```
honzajavorek: Should it be undefined or empty array? 🤔

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
```diff
@@ -33,16 +33,40 @@ function createTransactions(namespace, member, operation) {
...
+}
+
+function hrefFromParameters(path, queryParameters) {
+  const href = path.clone();
+  const queryString = queryParameters.keys().join(',');
```
honzajavorek: What if the parameter names need encoding before they become a part of the URI Template?

honzajavorek: E.g. `something[]`

kylef: That's not yet supported, unsupported characters will emit an error in the parameter parser so they won't reach this part of the code yet.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
kylef: It should be undefined because then we won't set an empty array in the `attributes` for `hrefVariables`. If we do set undefined value in attributes it won't turn up in the result.

The line further down that makes use will behave as described:

```js
transition.hrefVariables = hrefVariablesFromParameters(namespace, parameters);
```

NOTE, if we do set an empty array when there are no parameters, we would then be overriding the parameters section of the path item in this particular operation which is not desired.

kylef: Note, `transition.hrefVariables = undefined` will translate to a `transition.attributes.set('hrefVariables', undefined)`

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
honzajavorek: ok, didn't see FIXME, but maybe there's an issue already

PR #120, opened 2019-01-28
Support unreserved characters in parameter name

kylef:
We can support additional characters in a parameter name (`-`, `.`, `_`, and `~`) as per URI Template support.


packages/fury-adapter-oas3-parser/lib/parser/oas/parseParameterObject.js
```diff
@@ -27,9 +27,9 @@ const isValidInValue = R.anyPass([
   hasValue('cookie'),
 ]);
 
-
+const unreservedCharacterRegex = /^[A-z0-9\\.\\_\\~\\-]+$/;
```
kylef: I've moved out the creation of the regex expression outside of the function for performance reasons. We only have to compile the regex expression once when the parser is loaded instead of for each parameter.

kylef:
I've slipped in an additional commit(s) to improve the annotation message format for consistency. The format for unsupported messages is usually ${name} ${key} and then why its unsupported.

PR #121, opened 2019-01-28
Don't emit errors for missing path parameters

kylef:
Since this error doesn't count for both path item and operation parameters it can produce false positives causing a document to be unparsable (as it incorrectly causes an error).

We should validate that the parameters contain all the path variables. For now I am going to remove the validation logic as it is currently doing more harm than good to OAS 3 user experiance. Upon merge I'll create an issue to track restoring this behaviour.


kylef:
I'd also like to note that Drafter doesn't create any warning/error in this case either.

PR #122, opened 2019-01-28
Don't expose error that reference param wasn't defined when it was

kylef:
In some cases you can reference a parameter that couldn't be parsed (such as if it was an unsupported cookie parameter). This would cause the dereferencing to fail.

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
@@ -16,6 +16,53 @@ const name = 'Components Object';
 const unsupportedKeys = ['responses', 'examples', 'requestBodies', 'headers', 'securitySchemes', 'links', 'callbacks'];
 const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
 
+const valueIsObject = R.compose(isObject, getValue);
+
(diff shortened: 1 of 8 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
```diff
@@ -48,7 +48,16 @@ function parseComponentsObject(context, element) {
     const Member = R.constructN(2, namespace.elements.Member)(member.key);
     const parseResult = parseParameterObject(context, member.value);
     // Wrap non-annotation elements in member element
-    return R.map(R.unless(isAnnotation, Member), parseResult);
+    const result = R.map(R.unless(isAnnotation, Member), parseResult);
```
honzajavorek: It took me a while to read this, maybe because I was a bit out of context in the beginning. I'd at least rename `result` to something more descriptive, perhaps `parseResultMembers`.

kylef: Good point, I've refactored significantly to try and address this. Please review https://github.com/apiaryio/api-elements.js/pull/122/commits/f80dd3f0afb8a699a6b5f1bbd0109cdf4d921cb3 and I'll follow up with any feedback separately.

PR #127, opened 2019-01-29
Support referencing responses

kylef:


packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
```diff
@@ -48,9 +49,10 @@ const parseComponentMember = R.curry((context, parser, member) => {
   // Create a Member Element with `member.key` as the key
   const Member = R.constructN(2, context.namespace.elements.Member)(member.key);
 
-  const parseResult = pipeParseResult(context.namespace,
-    parser(context),
-    Member)(member.value);
+  const parseResult = R.map(
```
kylef: NOTE, this minor refactoring of this code is required to support referencing a component that has multiple values. For example a "response object" in OAS 3 maps to multiple HttpResponse elements. The map is simpler than previous implementation where we are mapping (wrapping) all non annotation elements in members.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseResponsesObject.js
```diff
@@ -31,8 +34,26 @@ const isResponseField = R.anyPass([isStatusCode, isStatusCodeRange, hasKey('defa
 function parseResponsesObject(context, element) {
   const { namespace } = context;
 
+  const createUnsupportedStatusCodeWarning = (member) => {
+    let message;
+
+    if (member.key.toValue() === 'default') {
+      message = `'${name}' default responses unsupported`;
```
honzajavorek:
```suggestion
      message = `'${name}' default responses are unsupported`;
```

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseResponseObject-test.js
```diff
@@ -12,38 +12,19 @@ describe('Response Object', () => {
...
-  describe('status code', () => {
-    it('provides a warning for default responses', () => {
-      const response = new namespace.elements.Member('default', {});
-      const result = parse(context, response);
-
-      expect(result.length).to.equal(1);
-      expect(result).to.contain.warning("'Response Object' default responses unsupported");
-    });
-
-    it('provides a warning for range responses', () => {
```
honzajavorek: This test was removed, but I don't see a replacement

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseResponseObject-test.js
```diff
@@ -12,38 +12,19 @@ describe('Response Object', () => {
...
     });
     const result = parse(context, response);
 
     expect(result.length).to.equal(1);
     const httpResponse = result.get(0);
     expect(httpResponse).to.be.instanceof(namespace.elements.HttpResponse);
-    expect(httpResponse.attributes.get('statusCode').toValue()).to.be.equal('200');
-  });
-
-  describe('status code', () => {
```
honzajavorek: This test was removed, but I don't see a replacement

kylef: There is already a test for this behaviour in parseResponsesObject-test.js, you can see it in the diff below.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseResponseObject-test.js
kylef: There is already a test for this behaviour in parseResponsesObject-test.js, you can see it in the diff below.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseResponsesObject.js
kylef: Added in 41cbb86

PR #129, opened 2019-01-29
Support required property in schema object

kylef:


packages/fury-adapter-oas3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -58,10 +58,24 @@ function parseSchema(context) {
   const parseSubSchema = element => parseReference('schemas', R.uncurryN(2, parseSchema), context, element, true);
   const parseProperties = parseObject(context, `${name}' 'properties`, R.compose(parseSubSchema, getValue));
 
+  const arrayElementToParseResult = array => new namespace.elements.ParseResult(array.content);
+  const parseArray = itemParser => pipeParseResult(namespace,
```
kylef: I wrote this in a way so we can extract it into a standalone component "parseArray" in the future

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -58,10 +58,24 @@ function parseSchema(context) {
...
+    R.unless(isArray, createWarning(namespace, `'${name}' 'required' is not an array`)),
+    R.compose(
```
kylef: We're converting the array element into a parse result of elements from the array and then mapping those elements into a new parse result using the item parser. The item parser either returns an element or it returns an annotation to go into the result.

Then, we have to convert the successful values into an array element below.

honzajavorek: I wonder whether comments like these shouldn't be comments in the code :)

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseSchemaObject-test.js
```diff
@@ -342,5 +342,58 @@ describe('Schema Object', () => {
...
+          type: 'object',
+          required: [1, true],
+        });
+        const result = parse(context, schema);
+
+        expect(result.length).to.equal(3);
+
+        expect(result.warnings.toValue()).to.deep.equal([
+          "'Schema Object' 'required' is not an array of strings",
+          "'Schema Object' 'required' is not an array of strings",
```
honzajavorek: The warning refers to the whole array, but is created for every array item, which feels/reads strange. Either it should be there only once for the whole array, or we could have warnings like...

Okay, I couldn't come up with a better warning 😅 Having sth like `item #2 of 'required'` feels overly detailed and doesn't read so clearly. At the same time, having just one warning for the whole array is an unnecessary loss of source map to each of the issues. If the `required` looks like this:

```
required:
  - yes
  - no
  - true
```

each warning will have a different source map with a different line, even though they have the same text. So I guess what you have here is the best compromise solution after all.

kylef: I have an idea, I'll update PR when I rebase it with new message

kylef: Updated to be: `'Schema Object' 'required' array value is not a string`. Note in cases like the following:

```
required:
  - name
  - id
  - null
```

You'd get a single warning about the last value only with specific source map.

PR #130, opened 2019-01-29
Support schema references from responses components

kylef:
Due to how the component parsers are integrated there would have been a bug referencing a schema from a responses component.

NOTE: This PR targets non-master as it depends on another PR so don't merge.


packages/fury-adapter-oas3-parser/lib/parser/parseReference.js
```diff
@@ -8,16 +8,12 @@ function isReferenceObject(element) {
...
 
-    if (component === 'schemas') {
-      const convertToReference = R.when(isDataStructure, (structure) => {
-        const element = new context.namespace.Element();
-        element.element = structure.content.id.toValue();
-        return new context.namespace.elements.DataStructure(element);
-      });
-
-      return R.map(convertToReference, result);
+    if (!returnReferenceElement && component === 'schemas') {
```
honzajavorek: Lines 11 and 13 together lead to a brain explosion. I'm used to brain explosions from reading core app code, but we shouldn't use explosions against the readers here.

The problem is that `returnReferenceElement || component === 'schemas'` and `!returnReferenceElement && component === 'schemas'` together is a logical puzzle, which hides the possible paths of the code. I need to compile them in my head and I'm very bad at that, usually forgetting about half of the combinations, i.e. paths which could happen.

With the previous PR, which added `returnReferenceElement`, I already thought whether it's not a smell, considering whether `parseReference` doesn't do way too many things and whether we couldn't break it down with some composition etc.

With all these conditions, I'm starting to be sure this needs some breaking down / polishing / structuring to be readable more clearly.

I think I can't give you a direct advice on what and how to improve the code - this comes as my gut feeling. Maybe after lunch 😄

packages/fury-adapter-oas3-parser/test/integration/components-test.js
```diff
@@ -23,4 +23,14 @@ describe('components', () => {
...
+
+  it("'Responses Object' response references with schema", () => {
+    const file = path.join(fixtures, 'responses-object-response-with-schema');
+    return testParseFixture(file);
+  });
+
+  it("'Schema Object' circular references", () => {
+    const file = path.join(fixtures, 'schema-object-circular');
+    return testParseFixture(file);
+  });
```
honzajavorek: Paranoia approved

packages/fury-adapter-oas3-parser/lib/parser/parseReference.js
kylef: I have some ideas on how to refactor it but I would like to get this PR functionality out. I've added a comment and renamed the variable so I think the logic is clearer already.

In the next few days I can prepare a follow up that refactors the schema parser further with its own referencing logic so it doesn't depend on parseReference. Oddly enough I already had that (and there was duplicate logic from parseReference and parseReferenceObject in parseSchemaObject, in efforts to deduplicate it ended up like this).

I think I can alter parseReferenceObject to not actually perform dereferencing, parseReference will handle actual dereferencing, and then I can have an alternative version of parseReference specific to schemas to address both problems (code duplication and complexity).

PR #132, opened 2019-01-30
Support request bodies

kylef:


packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
```diff
@@ -11,25 +13,46 @@ const parseObject = require('../parseObject');
...
-    request.method = method;
+  if (requests.length === 0) {
+    requests.push(new namespace.elements.HttpRequest());
+  }
+
+  const transactions = [];
 
-    return new namespace.elements.HttpTransaction([request, response]);
+  requests.forEach((request) => {
+    responses.forEach((response) => {
```
kylef: I went down an approach to refactor this to be a bit more functional, however I think the clarity of using nested forEach makes the code more easier to follow.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseReferenceObject.js
```diff
@@ -52,23 +52,22 @@ function parseReferenceObject(context, componentName, element, returnReferenceEl
...
+    const componentId = referenceParts[3];
+    if (!component.hasKey(componentId)) {
       return createError(namespace, `'${ref.toValue()}' is not defined`, ref);
     }
 
     if (returnReferenceElement) {
       const element = new context.namespace.elements.Element();
-      const id = referenceParts[3];
-      element.element = id;
+      element.element = componentId;
```
honzajavorek: 👍

packages/fury-adapter-oas3-parser/lib/parser/oas/parseReferenceObject.js
```diff
@@ -52,23 +52,22 @@ function parseReferenceObject(context, componentName, element, returnReferenceEl
...
 
-    if (element.value) {
-      return element.value;
-    }
-
-    return new namespace.elements.ParseResult([]);
+    return new namespace.elements.ParseResult(
+      component
+        .filter((value, key) => key.toValue() === componentId && value)
+        .map(value => value)
```
honzajavorek: What is the reason for this line?

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseOperationObject-test.js
```diff
@@ -389,4 +378,88 @@ describe('Operation Object', () => {
       });
     });
   });
+
+  describe('#responses', () => {
```
kylef: To be honest, this specific test case has nothing to do with this PR. Except I was debugging some behaviour related and noticed there wasn't a test for responses.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseComponentsObject-test.js
```diff
@@ -156,25 +156,61 @@ describe('Components Object', () => {
     });
   });
 
-  describe('warnings for unsupported properties', () => {
-    it('provides warning for unsupported examples key', () => {
+  describe('#requestBodies', () => {
+    it('provides a warning when responses is non-object', () => {
```
honzajavorek: responses?

packages/fury-adapter-oas3-parser/lib/parser/oas/parseReferenceObject.js
kylef: I spent some fun fighting the differences between filter/map on an ObjectSlice(ObjectElement) while writing this code. Filter here is returning an ObjectSlice of Member Elements, however the map transform function signature is expanded to be (`value`, `key`, `member`) and thus I want to map the member to value, value is the first argument so it is just passing it back through.

honzajavorek: I was suspecting something like this, but better asked :)

PR #139, opened 2019-02-04
Add support for description, default and example schema object properties

kylef:


packages/fury-adapter-oas3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -21,12 +21,10 @@ const unsupportedKeys = [
   'minItems', 'uniqueItems', 'maxProperties', 'minProperties', 'required',
 
   // JSON Schema + OAS 3 specific rules
-  'allOf', 'oneOf', 'anyOf', 'not', 'additionalProperties', 'description',
-  'format', 'default',
+  'allOf', 'oneOf', 'anyOf', 'not', 'additionalProperties', 'format',
 
   // OAS 3 specific
-  'nullable', 'discriminator', 'readOnly', 'writeOnly', 'xml', 'externalDocs',
```
kylef: `nullable` was incorrectly listed here.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseSchemaObject-test.js
```diff
@@ -533,4 +533,202 @@ describe('Schema Object', () => {
...
+      const parseResult = parse(context, schema);
+
+      expect(parseResult.length).to.equal(1);
+      expect(parseResult.get(0)).to.be.instanceof(namespace.elements.DataStructure);
+
+      const element = parseResult.get(0).content;
+      expect(element.attributes.get('default').toValue()).to.equal('my default');
+    });
+
+    it('allows a null default value when nullable is enabled', () => {
```
honzajavorek: I'd test null and enum interplay as well, it had some problems in the past IIRC

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseSchemaObject-test.js
```diff
@@ -533,4 +533,202 @@ describe('Schema Object', () => {
...
+        example: 'north',
+      });
+      const parseResult = parse(context, schema);
+
+      expect(parseResult.length).to.equal(2);
+      expect(parseResult).to.contain.warning(
+        "'Schema Object' 'example' is not included in 'enum'"
+      );
+    });
+  });
```
honzajavorek: If having both default and example, will it validate both?

kylef: Yes, but I can add a test.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -75,6 +73,63 @@ function constructArrayStructure(namespace, schema) {
...
+  return value.element === expectedElementType;
+};
+
+// Returns whether the given element value matches an enumeration of fixed values
+const valueMatchesEnumerationValues = (enumeration, value) => {
+  const permittedValues = enumeration.attributes.getValue('enumerations');
+  return permittedValues.includes(value.toValue());
+};
+
+function validateValuesMatchSchema(context, object) {
```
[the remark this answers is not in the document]
kylef: Yes, the difference is there is that MSON is the way you delcare data structures. We are not doing anything with schema, i.e, we don't compute data structure from schema or body from schema.

Perhaps in future we can take some of the logic out of this package and make it post-process for other adapters.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseSchemaObject-test.js
kylef: The test already uses an enum with nullable, perhaps I mis-understand the feedback can you please check if the schema below is what you desire.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSchemaObject.js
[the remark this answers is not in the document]
honzajavorek: Yup, I know it's different how it gets to API Elements, but from the user/developer perspective it's a very similar thing.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseSchemaObject-test.js
honzajavorek: You are right 🤦‍♂️ I was reading the test description.

So the test actually does what I wanted you to test, but the description doesn't indicate it, so enum is just an implementation detail of the test and not an explicit subject of the test. Someone could come in the future and change the test so it doesn't contain enum, but still tests a null default value when nullable is enabled.

For that reason I'd either do two tests, one with enum and one without, where the enum one explicitly mentions it i the test description, or just edit the description here to be explicit about testing the enum/nullable interplay as well.

Does it make sense what I'm writing?

kylef: Certainly does, will add additional tests. There are actually three cases to test:

- when there is no type or enum
- there is a type
- there is an enum

PR #142, opened 2019-02-05
Never return an error and a result

kylef:
BREAKING CHANGE: Asynchronous callbacks will never be called with an error and a parse result anymore.

This changeset aligns with how callbacks work in general in Node, you would never usually expect to see both a result (in our case a parse result element) and an error in a returned callback (https://nodejs.org/dist/latest-v10.x/docs/api/errors.html#errors_error_first_callbacks). This breaks behaviours such as promises and async/await. We do regard a parsing a problematic document (such as one with invalid API Blueprint syntax or one that does not meet validation requirements) a success in terms that we could parse the document and we have returned a valid parse result. Not to mention, the error property of the callback for some of our parsers may not even be an instance of `Error` but an annotation. Going forward, the error field would be used for actual failures to parse the document, for example Drafter had problems allocating memory or some system API failed us and we was not able to parse the document. Another example would be with a remote Fury adapter which hits a HTTP parsing API where the connection could fail, the attempt to parse the document failed.

In past we've had to work around these behaviours, such as in https://gist.github.com/kylef/aea36b5992acac4a3782981393e9f998#file-fury-server-js-L11-L30.

### DIFF
--- a/packages/fury/lib/fury.js
+++ b/packages/fury/lib/fury.js
@@ -57,14 +57,17 @@ class Fury {
     const adapter = this.findAdapter(source, mediaType, 'validate');
 
     if (!adapter) {
-      return this.parse({ source, mediaType, adapterOptions }, (err, result) => {
-        if (result && result.annotations.length > 0) {
-          const { ParseResult } = this.minim.elements;
-          const parseResult = new ParseResult(result.annotations);
-          done(err, parseResult);
-        } else {
-          done(err, null);
+      return this.parse({ source, mediaType, adapterOptions }, (err, parseResult) => {
+        if (err) {
+          return done(err);
         }
+
+        if (parseResult.annotations.length > 0) {
+          const { ParseResult } = this.minim.elements;
+          return done(null, new ParseResult(parseResult.annotations));
+        }
+
+        return done(null, null);
       });
     }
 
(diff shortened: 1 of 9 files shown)


packages/fury/lib/fury.js
```diff
@@ -75,12 +74,16 @@ class Fury {
       options = Object.assign(options, adapterOptions);
     }
 
-    return adapter.validate(options, (err, elements) => {
-      if (!elements || elements instanceof this.minim.Element) {
-        done(err, elements);
-      } else {
-        done(err, this.load(elements));
+    return adapter.validate(options, (err, parseResult) => {
+      if (!parseResult) {
```
pksunkara: Shouldn't this be `if (err) {`?

packages/fury/lib/fury.js
```diff
@@ -106,12 +109,16 @@ class Fury {
         options = Object.assign(options, adapterOptions);
       }
 
-      return adapter.parse(options, (err, elements) => {
-        if (!elements || elements instanceof this.minim.Element) {
-          done(err, elements);
-        } else {
-          done(err, this.load(elements));
+      return adapter.parse(options, (err, parseResult) => {
+        if (!parseResult) {
```
pksunkara: Shouldn't this be `if (err) {`?

packages/fury/lib/fury.js
```diff
@@ -58,11 +58,10 @@ class Fury {
...
-      return this.parse({ source, mediaType, adapterOptions }, (err, result) => {
-        if (result && result.annotations.length > 0) {
+      return this.parse({ source, mediaType, adapterOptions }, (err, parseResult) => {
+        if (parseResult && parseResult.annotations.length > 0) {
           const { ParseResult } = this.minim.elements;
-          const parseResult = new ParseResult(result.annotations);
-          done(err, parseResult);
+          done(null, new ParseResult(parseResult.annotations));
         } else {
           done(err, null);
```
pksunkara: Move this up and use `if (err) {`

packages/fury/lib/fury.js
kylef: I was protecting against older adapters which behave incorrectly, if we have an adapter that returns both an error and a parseResult, fury will discard the error

honzajavorek:
Good! I believe this allows to simplify https://github.com/apiaryio/dredd-transactions as well

PR #146, opened 2019-02-05
valueOf method for Boolean, String, Number, Null

tjanc:
This PR introduces a `valueOf` method on Boolean, String, Number and Null elements.
Enum, Array, Object are not supported yet, but support will follow.

### Usage
The `valueOf` method interprets an API Element as a type (including all intricate details of the API Elements type system) and reduces it to a JSON value.

```
const element = new StringElement('hello');
const value = element.valueOf();
expect(value).to.equal('hello');
```
```
const element = new StringElement();
element.attributes.set('typeAttributes', new ArrayElement([
  new StringElement('nullable'),
]));
const value = element.valueOf();
expect(value).to.equal(null);
```
An API Element can annotate specific values int their value category as `content`, `sample` or `default`. Also, some Elements have trivial values that can be generated (e.g. the empty string for String Element). 
For that reason, an option `{source: true}` MAY be passed to `valueOf` to also return, in pair, the source of the value:
```
const element = new StringElement();
element.attributes.set('default', new StringElement('moin'));
const value = element.valueOf({ source: true });
expect(value).to.deep.equal(['moin', 'default']);
```
```
const element = new StringElement();
const value = element.valueOf({ source: true });
expect(value).to.deep.equal(['', 'generated']);
```
The second member of the returned pair is one of `'content', 'sample', 'default', 'generated'`
  - `content` refers to the Element content
  - `sample` refers to Element samples
  - `default` refers to the Element default
  - `nullable` refers to the nullable type attribute (i.e. no content, no samples, no default, but the Element is nullable)
  - `generated` refers to a default based on the Element type (e.g. the empty string `''` can be a generated value for an empty String Element lacking content, sample, default and nullable)

***

Edit 1: modified interface from `.instantiate()` to `.valueOf({ source: true })`, also supporting `.valueOf()`
Edit 2: added `'nullable'` source type


char0n:
Can we pls change the name of `instantiate` to something more understandable ?
Maybe If you explain the logic for naming it here, I'd understand, but now it's a little bit confusing for me. 

And thank you for working on this! ;]

tjanc:
Modifying the interface to match `const [value, source] = element.valueOf({ source: true });`, `const value = element.valueOf();`. Thanks char0n, kylef.

honzajavorek:
This is cool, it could remove quite some code from https://github.com/apiaryio/dredd-transactions

packages/minim-api-description/lib/api-description.js
```diff
@@ -34,8 +34,10 @@ const copy = require('./elements/copy');
 const category = require('./elements/category');
 const extension = require('./elements/extension');
 const enumeration = require('./elements/enum');
+const definevalueOf = require('./define-value-of');
```
honzajavorek: camel case? defineValueOf

packages/minim-api-description/test/value-of-test.js
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  it('generates null if no content, default, samples and nullable', () => {
+    const element = new StringElement();
+    element.attributes.set('typeAttributes', new ArrayElement([
+      new StringElement('nullable'),
+    ]));
+    const value = element.valueOf({ source: true });
+
+    expect(value).to.deep.equal([null, 'content']);
```
pksunkara: Shouldn't this be `[null, 'nullable']`?

char0n:
Looks good. Thanks

packages/minim-api-description/lib/api-description.js
tjanc: how did you notice that? :smiley:

packages/minim-api-description/lib/define-instantiate.js
```diff
@@ -0,0 +1,120 @@
...
+        }
+      }
+    }
+    return null;
+  }
+
+  const isPrimitive = e => (e instanceof StringElement) || (e instanceof NumberElement) || (e instanceof BooleanElement);
+
+  // Generate default value for leaf elements (Boolean, String, Number, Null)
+  function generateValue(e) {
```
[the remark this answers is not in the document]
tjanc: Is `generateTrivialValue` is better? `generateDefaultValue` could mistakenly suggest interpretation of the `default` attribute.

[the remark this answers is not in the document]
kylef: Yes I agree, that's a better idea I forgot that default would become an overloaded term in this area of the code.

PR #150, opened 2019-02-07
valueOf method for EnumElement

tjanc:
Adds support for the method introduced in PR #146 on Enum Element.

### DIFF
--- a/packages/minim-api-description/lib/define-value-of.js
+++ b/packages/minim-api-description/lib/define-value-of.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-bitwise, no-underscore-dangle */
 
 const setFlag = (mask, options) => (options | mask);
+const clearFlag = (mask, options) => (options & ~mask);
 const isFlag = (mask, options) => (options & mask) !== 0;
 
 const FIXED_FLAG = 1 << 0;
@@ -56,6 +57,7 @@ module.exports = (namespace) => {
     Boolean: BooleanElement,
     Number: NumberElement,
     Null: NullElement,
+    Enum: EnumElement,
   } = namespace.elements;
 
   function findFirstSample(e) {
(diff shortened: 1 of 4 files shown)


packages/minim-api-description/lib/define-value-of.js
```diff
@@ -88,33 +90,45 @@ module.exports = (namespace) => {
...
-      return wrapResult(dflt.content, 'default', opts);
+      return wrapResult(recurse(dflt.content, clearFlag(SOURCE_FLAG, opts)), 'default', opts);
     }
     if (isFlag(NULLABLE_FLAG, opts)) {
       return wrapResult(null, 'nullable', opts);
     }
+    if (isEnumElement(e)) {
+      const enums = e.enumerations;
+      if (enums && enums.content && enums.content[0]) {
+        return wrapResult(recurse(enums.content[0], clearFlag(SOURCE_FLAG, opts)), 'generated', opts);
```
pksunkara:
```suggestion
        return wrapResult(recurse(enums.content[0], clearFlag(SOURCE_FLAG, opts)), 'content', opts);
```

packages/minim-api-description/test/value-of-test.js
```diff
@@ -413,3 +414,212 @@ describe('valueOf StringElement with source', () => {
...
+  it('prefers a sample over a default', () => {
+    const element = new EnumElement();
+    element.enumerations = new ArrayElement([
+      new BooleanElement(),
+      new NumberElement(),
+      new StringElement(),
+    ]);
+    element.attributes.set('default', new EnumElement('moin'));
+    element.attributes.set('samples', new ArrayElement([
+      new EnumElement(new EnumElement('zdravicko')),
```
pksunkara: Why is this double EnumElement?

packages/minim-api-description/test/value-of-test.js
```diff
@@ -413,3 +414,212 @@ describe('valueOf StringElement with source', () => {
...
+  it('prefers content over default and samples', () => {
+    const element = new EnumElement('hello');
+    element.enumerations = new ArrayElement([
+      new BooleanElement(),
+      new NumberElement(),
+      new StringElement(),
+    ]);
+    element.attributes.set('default', new EnumElement('moin'));
+    element.attributes.set('samples', new ArrayElement(
+      new EnumElement(new EnumElement('zdravicko'))
```
pksunkara: Why is this double EnumElement?

packages/minim-api-description/lib/define-value-of.js
pksunkara: I am probably wrong on this. Not sure. We might want to forward the value from the enumeration to here, right?

tjanc: `generated` is correct here; `enumerations` describes the type and taking the first of those _is_ generating a value.

packages/minim-api-description/test/value-of-test.js
tjanc: To test whether recursion works.

packages/minim-api-description/test/value-of-test.js
tjanc: To test whether recursion works.

PR #155, opened 2019-02-11
chore(project): upgrade dependencies

pksunkara:


honzajavorek:
Tests are failing

pksunkara:
@honzajavorek Updated.

packages/fury-adapter-oas3-parser/package.json
```diff
@@ -19,18 +19,18 @@
     "test:integration:fix": "env GENERATE=1 mocha test/integration"
   },
   "dependencies": {
-    "media-typer": "1.0.1",
+    "media-typer": "^1.0.1",
     "ramda": "0.26.1",
-    "yaml-js": "0.2.3"
+    "yaml-js": "^0.2.3"
```
honzajavorek: Why do you have inconsistent approach to having `^` in the version number? I'd suggest to remove it everywhere and to have Greenkeeper with explicit upgrades instead. This way ramda is pinned, but yaml-js isn't, even though they're both with 0 major version. What's the reasoning behind this?

pksunkara: We are being consistent about having `^`. There are only a few deps where we are making an exception because of previous experiences. Ramda here doesn't have `^` yet because I wanted to wait for Kyle to come back so he can review that change.

honzajavorek: Previously @char0n pointed out that Ramda should be pinned, because they release breaking changes sometimes. By Semantic Versioning, you get no guarantees when the major version is 0. I don't see how that doesn't apply for yaml-js as well.

Your changes make everything to use `^`, while previously nothing in the OAS3 parser was using it except of `eslint`, which could have been a mistake.

I don't know how to decide here, IMHO it's up to @kylef to make it clear, but he's on vacation now and I don't want to block you. I think the best solution would be to bring another reviewer in to approve or reject the PR.

pksunkara: The `^` I added, those same packages with those same version with those same `^` were being used elsewhere in the monorepo. Since we are not using `^`, the dependencies were getting installed in a non-monorepo way. So, I fixed it.

honzajavorek: OK, makes sense. Thanks and sorry for holding this back.

PR #165, opened 2019-02-20
valueOf method for Array & Object Elements

tjanc:
Adds support for the method introduced in PR #146 on Array Element and Object Element.


pksunkara:
Can you add a changelog entry? Thanks.

kylef:
I think the current master changelog already covers it:

> Introduced .valueOf method reducing an API Element from its type to a normalised value.

https://github.com/apiaryio/api-elements.js/blob/8d4117f76f1d9d742b023037a0666d17f3dc72ff/packages/minim-api-description/CHANGELOG.md#enhancements

PR #176, opened 2019-03-04
Support referencing properties in data structures

### LINKED ISSUE #173: Support for data structure generation when referencing property from another schema
Given the following OpenAPI 2 document which contains a reference to another definition's property (`$ref: '#/definitions/User/properties/name'`) the OpenAPI 2 parser is not emitting a data structure for the structure that uses the reference:

```yaml
swagger: '2.0'
info:
  title: Example
  version: '1.0.0'
paths: {}
definitions:
  User:
    type: object
    properties:
      name:
        type: string
  Company:
    type: object
    properties:
      name:
        $ref: '#/definitions/User/properties/name'

```

#### Expected result

The parse result contains two data structures, "User" and "Company".

```
$ npx fury example.yaml | jq -r .content[0].content[0].content[].content.meta.id.content
definitions/User
definitions/Company
```

#### Actual Result

Results in a parse result which contains a single data structure "User", there is no structure present for "Company" definition:

```
$ npx fury example.yaml | jq -r .content[0].content[0].content[].content.meta.id.content
definitions/User
```

---

The cause of this is because a property is not a data structure in API Elements and we cannot reference the arbitary location so the parser is stripping the particular structure out of its response.

This is unfortunately not so trivial to fix and I think there are two possibilities:

1) We de-reference the property in this particular case and expose it in the "Company" data structure without any references.
1) We derive a reusable data-structure from the property and make a reference against it. So the parse result will contain 3 data structures, "User", "Company" and "User.name" which they both reference.

kylef:
Closes #173

### DIFF
--- a/packages/fury-adapter-swagger/test/schema-test.js
+++ b/packages/fury-adapter-swagger/test/schema-test.js
@@ -5,6 +5,7 @@ const { DataStructureGenerator } = require('../lib/schema');
 const namespace = new Fury().minim;
 
 const {
+  Element,
   String: StringElement,
   Number: NumberElement,
   Boolean: BooleanElement,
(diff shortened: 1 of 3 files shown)


packages/fury-adapter-swagger/test/schema-test.js
```diff
@@ -943,4 +944,79 @@ describe('JSON Schema to Data Structure', () => {
...
+      const dataStructure = schemaToDataStructure(schema, root);
+
+      expect(dataStructure.element).to.equal('dataStructure');
+      expect(dataStructure.content).to.be.instanceof(ObjectElement);
+
+      const user = dataStructure.content.get('user');
+      expect(user.element).to.equal('definitions/User');
+    });
+
+    it('produces element from object containing reference to another element property', () => {
```
kylef: The following test is for the new behaviour, the other two tests are increasing related test coverage.

packages/fury-adapter-swagger/test/schema-test.js
```diff
@@ -5,6 +5,7 @@ const { DataStructureGenerator } = require('../lib/schema');
 const namespace = new Fury().minim;
 
 const {
+  Element,
   String: StringElement,
   Number: NumberElement,
   Boolean: BooleanElement,
```
honzajavorek: Side note, I wouldn't rename the classes like this, not even just for the purpose of the tests. It's not that unusual one needs to use JavaScript's builtin `String`, `Number`, etc. types, and this makes it impossible. Reading the tests can also be confusing (_is this a JavaScript's String or is this a minim element?_)

kylef: We're renaming the elements from what is stored in the namespaces `elements` collection. The namespace elements collection is named as they are in Refract Spec "Number", "String", "Boolean", "HttpRequest", "Resource" etc. As these names clash in JavaScript land we're calling them `NumberElement`, `StringElement` etc. JavaScript doesn't have a namespace system, in C++ or such this is clearer because it can be such as `refract::String` where `refract` is a namespace for the structures.

kylef: I find the JS syntax for this a little hard to remember as I think of it as backwards. It's effectively:

```js
const StringElement = minim.elements['String'];
```

PR #177, opened 2019-03-04
refactor(oas3): added parseArray predicate

pksunkara:


packages/fury-adapter-oas3-parser/lib/parser/parseArray.js
```diff
@@ -0,0 +1,96 @@
...
+  // the result to a parse result if it isn't already a parse result.
+  const transform = R.pipe(parseValue, coerceParseResult);
+
+  /**
+   * Converts the given parse result of values into parse result of an array
+   * @param parseResult {ParseResult<Element>}
+   * @returns ParseResult<Array>
+   */
+  const convertParseResultMembersToArray = (parseResult) => {
+    const values = R.filter(isNotAnnotation, parseResult);
```
kylef:
```suggestion
    const values = R.reject(isAnnotation, parseResult);
```

There is also [`R.reject`](https://ramdajs.com/docs/#reject) which complients `R.filter` so we don't have to have negative predicates.

packages/fury-adapter-oas3-parser/lib/parser/parseArray.js
```diff
@@ -0,0 +1,96 @@
...
+ * transformation and then reconstructs a parse result containing all
+ * of the results.
+ *
+ * |------->------->------------------->--------------------|
+ * |       > value > parseValue(value) >                    |
+ * | Array > value > parseValue(value) > ParseResult<Array> |
+ * |       > value > parseValue(value) >                    |
+ * |------->------->------------------->--------------------|
+ *
+ * @param namespace
```
kylef: The signature of the function is `context` not namespace.

```suggestion
 * @param context
```

PR #178, opened 2019-03-04
Don't duplicate property descriptions on member

kylef:
Spotted by Standa, previous we have been duplicating descriptions on the member element when you have a property that contains a description:

```yaml
definitions:
  User:
    type: object
    properties:
      name:
        type: string
        description: a name
```

The previous logic was an attempt to "move" the description from a property from the member to align with other parsers. However due to language differences I think that this does not make sense because in JSON Schema the description is apart of the type. Especially when it comes to referencing:

```yaml
definitions:
  User:
    type: object
    properties:
      name:
        $ref: '#/definitions/name'
  name:
    type: string
    description: a name
```


pksunkara:
@kylef conflicts

PR #180, opened 2019-03-05
impl(remote): initial commit of configurable "remote" parser

### LINKED ISSUE #143: Provide a "remote" Fury adapter
As we provide an API (`https://api.apielements.org`) which can handle validation, parsing and serialisation we should offer this as an adapter itself so users can drop-in replace the adapters with remote parsing. It should be called `fury-adapter-remote`.

```js
const { Fury } = require('fury');
const FuryRemoteAdapter = require('fury-adapter-remote');

const fury = new Fury();
fury.use(new FuryRemoteAdapter({
  // the default value, but consumers should be able to override to use their own deployment
  url: 'https://api.apielements.org',

  // the collection of "parse", media types we want this
  // instance of the adapter to handle.
  // NOTE, this allows you to use the API for one media type but
  // another local adapter for another.
  parseMediaTypes: ['text/vnd.apiblueprint', 'application/oai.oas'],

  // the collection of "serialize", media types we want this
  // instance of the adapter to handle.
  serializeMediaTypes: ['text/vnd.apiblueprint'],
}));
```

```js
fury.validate(...);
fury.parse(...);
fury.serialise(...);
```

I'd suggest that the remote adapter use [`axios`](https://github.com/axios/axios) for talking to the API since is both Node and browser friendly and we're using it elsewhere. The adapter should perform POST requests to our public endpoints such as `/validate`, `/parse` and `/compose` using API Elements 1.0 content-type.

klokane:
Inital commit.

There is a few problematic parts:
- `Fury` actually does not provide way to distinguish among individual MediaTypes related to operations, so implementation returns MediaType for all types of operation
- similar problem is, there is not MediaType passed into individual operations. So we need to detect it again in adapter, but in general it can differ from MediaType specified by user.

But I mean it not should be part of this PR, it should be introduced by individual commit and changes in `remote adapter` should follow it.

Closes #143

### DIFF
--- a/packages/fury-adapter-remote/README.md
+++ b/packages/fury-adapter-remote/README.md
@@ -3,14 +3,13 @@
 [![NPM version](https://img.shields.io/npm/v/fury-adapter-remote.svg)](https://www.npmjs.org/package/fury-adapter-remote)
 [![License](https://img.shields.io/npm/l/fury-adapter-remote.svg)](https://www.npmjs.org/package/fury-adapter-remote)
 
-This adapter provides support for parsing, validation and serialization via [Public API Service](http://api.apielements.org/) in [Fury](https://github.com/apiaryio/api-elements.js/tree/master/packages/fury).
+This adapter provides support for parsing, validation and serialization via the [API Elements Web Service](https://api.apielements.org/) in [Fury](https://github.com/apiaryio/api-elements.js/tree/master/packages/fury).
 
-In adapter is configurable and can be configured to work with any public API service compatible with [API BlueprintAPI Service](https://apiblueprintapi.docs.apiary.io)
-It provide default configuration connected to [Public API Service](http://api.apielements.org/) to working "out of the box"
+The remote adapter can be configured to work with any API service compatible with [API Blueprint API](https://apiblueprintapi.docs.apiary.io).
+By default the [API Elements Web Service](http://api.apielements.org/) is used.
 
 ```js
 const defaultOptions = {
-  // the default value, but consumers should be able to override to use their own deployment
   url: 'https://api.apielements.org',
 
   parseEndpoint: '/parser',
@@ -45,7 +44,7 @@ const defaultOptions = {
 ## Install
 
 ```sh
-npm install fury-adapter-remote
+$ npm install fury-adapter-remote
 ```
 
 ## Usage
@@ -55,10 +54,9 @@ const { Fury } = require('fury');
 const FuryRemoteAdapter = require('fury-adapter-remote');
 
 const fury = new Fury();
-fury.use(new FuryRemoteAdapter({});
+fury.use(new FuryRemoteAdapter(options);
 
 fury.validate(...);
 fury.parse(...);
 fury.serialise(...);
 ```
-
--- a/packages/fury-adapter-remote/README.md
+++ b/packages/fury-adapter-remote/README.md
@@ -1,0 +1,64 @@
+# Fury Remote Adapter
+
+[![NPM version](https://img.shields.io/npm/v/fury-adapter-remote.svg)](https://www.npmjs.org/package/fury-adapter-remote)
+[![License](https://img.shields.io/npm/l/fury-adapter-remote.svg)](https://www.npmjs.org/package/fury-adapter-remote)
+
+This adapter provides support for parsing, validation and serialization via [Public API Service](http://api.apielements.org/) in [Fury](https://github.com/apiaryio/api-elements.js/tree/master/packages/fury).
+
+In adpter is cofigurable and can be cofigured to work with any public API service compatible with [API BlueprintAPI Service](https://apiblueprintapi.docs.apiary.io)
+It provide default configuration connected to [Public API Service](http://api.apielements.org/) to working "out of the box"
+
+```js
+const defaultOptions = {
+  // the default value, but consumers should be able to override to use their own deployment
+  url: 'https://api.apielements.org',
+
+  parseEndpoint: '/parser',
+  validateEndpoint: '/validate',
+  serializeEndpoint: '/composer',
+
+  // the collection of "parse", media types we want this
+  // instance of the adapter to handle.
+  // NOTE, this allows you to use the API for one media type but
+  // another local adapter for another.
+  parseMediaTypes: [
+    'text/vnd.apiblueprint',
+    'application/swagger+json',
+    'application/swagger+yaml',
+    'application/vnd.oai.openapi',
+    'application/vnd.oai.openapi+json',
+  ],
+
+  // the collection of "serialize", media types we want this
+  // instance of the adapter to handle.
+  serializeMediaTypes: [
+    'application/vnd.refract+json',
+    'application/vnd.refract.parse-result+json',
+  ],
+
+  // fallback to try send input, if not indentified by deckardcain
+  defaultParseMediaType: 'text/vnd.apiblueprint',
+  defaultSerializeMediaType: 'application/vnd.refract+json',
+};
+```
+
+## Install
+
+```sh
+npm install fury-adapter-remote
+```
+
+## Usage
+
+```js
+const { Fury } = require('fury');
+const FuryRemoteAdapter = require('fury-adapter-remote');
+
+const fury = new Fury();
+fury.use(new FuryRemoteAdapter({});
+
+fury.validate(...);
+fury.parse(...);
+fury.serialise(...);
+```
+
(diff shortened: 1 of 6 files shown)


packages/fury-adapter-remote/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "lint:fix": "eslint . --fix",
+    "test": "mocha test"
+  },
+  "engines": {
+    "node": ">=4"
+  },
+  "dependencies": {
+    "axios": "^0.18.0",
+    "deckardcain": "^0.4.1",
+    "fury": "^3.0.0-beta.8"
```
kylef: This should become a peerDependency and a devDependency, check how they are defined in other adapters (`fury-adapter-oas3-parser/package.json`).

Since a user of the adapter is expected to depend on Fury, having a hard dependency will cause a duplicate dependency.

packages/fury-adapter-remote/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "type": "git",
+    "url": "https://github.com/apiaryio/api-elements.js.git"
+  },
+  "scripts": {
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
+    "test": "mocha test"
+  },
+  "engines": {
+    "node": ">=4"
```
kylef: We're only testing on Node 6 onwards so let's make the range we support what we're testing in CI:

```suggestion
    "node": ">=6"
```

packages/fury-adapter-remote/test/fury-test.js
```diff
@@ -0,0 +1,216 @@
...
+        expect(result.length).to.be.equal(1);
+        expect(result.annotations.length).to.be.equal(1);
+        expect(result.annotations.get(0).content).to.be.equal('API version number must be a string (e.g. "1.0.0") not a number.');
+      });
+    });
+  });
+
+  describe('#serialize', () => {
+    it('ParseResult to blueprint', () => {
+      fury.serialize({ api: blueprintApi }, (err, result) => {
```
kylef: I don't think this test is correct given the API for Fury. `fury.serialize` accepts an "API" element,  (i,e instance of `Category`) and not a JSON Parse Result.

Take a look at the tests inside the API Blueprint Serializer at https://github.com/apiaryio/api-elements.js/blob/1cfc40a32e682d0fb236e5cd9ff12dded4088a44/packages/fury-adapter-apib-serializer/test/adapter-test.js#L32-L38 which demonstrate how the API is used.

Something like the following would work in the test where we are deserialising the JSON structure into fury elements:

```js
const api = fury.minim.serialiser.deserialise({
  element: 'category',
  meta: {
    classes: {
      element: 'array',
      content: [
        {
          element: 'string',
          content: 'api',
        },
      ],
    },
    title: {
      element: 'string',
      content: 'API',
    },
  },
  content: [],
});
```

This will uncover the implementation doesn't quite work for the expected input, you will need to update the logic to serialise the Category element:

```
  serialize({ api, minim }, cb) {
    const contentType = 'application/vnd.refract.parse-result+json; version=1.0';
    const content = minim.serialiser.serialise(api);

    axios({
      method: 'post',
      url: this.options.serializeEndpoint,
      baseURL: this.options.url,
      data: content,
      headers: {
        'Content-Type': contentType,
      },
```

packages/fury-adapter-remote/lib/adapter.js
```diff
@@ -0,0 +1,131 @@
...
+
+class FuryRemoteAdapter {
+  constructor(options) {
+    this.name = name;
+    this.options = options || defaultOptions;
+    this.mediaTypes = this.options.mediaTypes;
+  }
+
+  detect(source) {
+    return this.mediaTypes.indexOf(deckardcain.identify(source)) !== -1;
```
kylef: There's an [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) method on array which can be used:

```suggestion
    return this.mediaTypes.includes(deckardcain.identify(source));
```

packages/fury-adapter-remote/README.md
```diff
@@ -0,0 +1,63 @@
+
```
kylef:
```suggestion
```

Empty newline found at start.

packages/fury-adapter-remote/lib/adapter.js
```diff
@@ -0,0 +1,131 @@
...
+  parseEndpoint: '/parser',
+  validateEndpoint: '/validate',
+  serializeEndpoint: '/composer',
+
+  // the collection of "parse", media types we want this
+  // instance of the adapter to handle.
+  // NOTE, this allows you to use the API for one media type but
+  // another local adapter for another.
+
+  mediaTypes: [
```
kylef: The design in https://github.com/apiaryio/api-elements.js/issues/143#issue-406774120 separates the media types for parsing and serialisation, the reason for this is that you might want to configure the adapter to only parse API Blueprint but not support serializing API Blueprint so the media types would need to be separate arrays:

```
  // the collection of "parse", media types we want this
  // instance of the adapter to handle.

  // NOTE, this allows you to use the API for one media type but
  // another local adapter for another.
  parseMediaTypes: ['text/vnd.apiblueprint', 'application/oai.oas'],

  // the collection of "serialize", media types we want this
  // instance of the adapter to handle.
  serializeMediaTypes: ['text/vnd.apiblueprint'],
```

packages/fury-adapter-remote/lib/adapter.js
```diff
@@ -0,0 +1,131 @@
...
+    'text/vnd.apiblueprint',
+    'application/swagger+json',
+    'application/swagger+yaml',
```
kylef: Please add `application/vnd.oai.openapi` and `application/vnd.oai.openapi+json` to support OAS 3

kylef:
@klokane I'm all for working incrementally, if you want to get this PR merged sooner and make it easier to review you could remove the support for `serialize` and add it in a separate PR once addressing the feedback.

packages/fury-adapter-remote/test/fury-test.js
kylef: I think the interface can be unclear so I've documented it at https://github.com/apiaryio/api-elements.js/pull/183

packages/fury-adapter-remote/lib/adapter.js
klokane: I agree, but it is what i wrote in initial comment.
`detect()` is not able to check kind of operation you are asking for mediaType and same operation itself. If I remove it, it not will be able to serialisation later.

Maybe I can split it in for public space (to avoid changing interface in future), but internaly in `detect` I must do check for both.

```
return this.parseMediaTypes.includes(deckardcain.identify(source)) || serializeMediaTypes.includes(deckardcain.identify(source));
```

Seems to good solution for you?

packages/fury-adapter-remote/test/fury-test.js
klokane: I use plutonium documentation https://plutonium.docs.apiary.io/#reference/0/api-description-parser/compose where is written 

> API Elements is supported, you can send an API Elements API Category to the composer using the `application/vnd.refract+json` content type. You may also send an API Elements Parse Result as `application/vnd.refract.parse-result+json`.

So I understand both MediaTypes are valid as input and both are equivalent.

packages/fury-adapter-remote/package.json
klokane: ok

packages/fury-adapter-remote/package.json
klokane: will update, BTW on fury-adapter-oas3-parser` https://github.com/apiaryio/api-elements.js/blob/master/packages/fury-adapter-oas3-parser/package.json#L37

Is also `"node": ">=4"` is it correct?

kylef:
@klokane I've merged in https://github.com/apiaryio/api-elements.js/pull/184, want to rebase and update this PR ontop of it?

klokane:
rebased

klokane:
@kylef rebased

packages/fury-adapter-remote/README.md
```diff
@@ -0,0 +1,64 @@
+# Fury Remote Adapter
+
+[![NPM version](https://img.shields.io/npm/v/fury-adapter-remote.svg)](https://www.npmjs.org/package/fury-adapter-remote)
+[![License](https://img.shields.io/npm/l/fury-adapter-remote.svg)](https://www.npmjs.org/package/fury-adapter-remote)
+
+This adapter provides support for parsing, validation and serialization via [Public API Service](http://api.apielements.org/) in [Fury](https://github.com/apiaryio/api-elements.js/tree/master/packages/fury).
+
+In adpter is cofigurable and can be cofigured to work with any public API service compatible with [API BlueprintAPI Service](https://apiblueprintapi.docs.apiary.io)
```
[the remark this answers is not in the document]
klokane: fixed

packages/fury-adapter-remote/package.json
kylef: It's not, good catch. I can fix in separate PR.

PR #183, opened 2019-03-05
Add JSDoc comments to Fury

kylef:
I think it can be unclear what the public API to Fury, and the API between Fury and adapters actually is, so I have added documentation describing the properties and types. This isn't complete, it's to kick it off. I think next step we need to integrate it into our Sphinx docs. I have not really written much about any of the interface, this is more specifying the structures and types at the moment.

In doing so, I've updated existing jsdoc comments on internal interfaces to `@private` so they don't show up. Those developing can generate documentation exposing private interfaces. Perhaps we can host internal API docs too in Sphinx at a later date.

```shell
$ jsdoc --recurse packages/*/lib node_modules/minim/lib/
```

Using jsdoc I've generated the docs and put them online at https://apielements.rigidapp.com/Fury.html to make it easier to review.

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
@@ -25,6 +25,7 @@ const unsupportedKeys = ['servers', 'security', 'tags', 'externalDocs'];
  * @param member {MemberElement}
  * @returns {boolean}
  * @see unsupportedKeys
+ * @private
  */
 const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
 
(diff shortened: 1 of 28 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
```diff
@@ -25,6 +25,7 @@ const unsupportedKeys = ['servers', 'security', 'tags', 'externalDocs'];
  * @param member {MemberElement}
  * @returns {boolean}
  * @see unsupportedKeys
+ * @private
  */
 const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
```
char0n: Would it make sense for tacit (point-free) functions like these to add `@sig` containing Hindley-Millner type signature ? Functional programmers would welcome it when reading the code. Example of using it in jsdoc: https://github.com/char0n/ramda-adjunct/blob/master/src/allEqual.js#L16

How to write signatures: https://github.com/ramda/ramda/wiki/Type-Signatures

PR #186, opened 2019-03-12
Add support for json examples

alexylew:


kylef:
Hi @alexylew thanks for taking a look at this, it's exciting to see someone else contribute here.

I think we should break up this PR a little bit and first focus on getting "Media Type Object" parser adding the message body as you have. As a separate step I think we should evolve the schema object parser where needed. We should not create our own samples derived from the schema but instead the samples should represent what the user wrote in their OAS 3 document (example/examples defined in Schema Object). The logic for computing a JSON body should look deeper into the structure to compute an example which introspects the datastructure more as there is multiple places a value can come from (and not just samples). **The good news is we already have this algorithm in place**, the element will have a [`valueOf()` method](https://github.com/apiaryio/api-elements.js/blob/e3fe71b97b804f935ab27e95c0c5850ef2e3d480/packages/minim-api-description/lib/define-value-of.js#L212) which will compute a value looking in the default, samples, content of each element. valueOf can handle empty values, nullable values etc. I think we should evolve this method itself to look into references in the future (I'd tackle this separately and get the core support finished first).

I believe something like the following should suffice:

```js
const value = dataStructure.content.valueOf();
const body = JSON.stringify(value);
```

Then you should be able to remove the changes from the reference and schema parser. I'd like to tackle referencing a little differently, instead I think the code to compute the JSON body should deal with expansion/dereferencing when needed. That way we can handle cirlcular referencing and we can avoid unnessecery expansion. For example in the following schema:

```yaml
type: object
properties:
  user:
    $ref: '#/..some/ref'
example:
  user:
    name: kyle
```

We have an outter level example value and thus we do not need to expand the reference as that will just be additional computation that is not needed and can increase parsing times. I would consider (outside of this PR) altering `valueOf` to take an element tree and be able to dereference it when traversing the structure. We can discuss how this can work in the future once we've got the base support complete.

We should only attempt to create a JSON body if the content-type in the media object contains 'application/json' or similiar content type. You can find logic in `parseMediaTypeObject.js` inside `parseExample` which contains `isJSONMediaType`. We can extract this method out and reuse it. We should only generate our example IF the messageBody on line 84 didn't exist, we don't want to compute another message body if the user already supplied one in their OAS 3 document. We can do something like:

```js
if (!messageBody && dataStructure && isJSONMediaType(mediaType)) {
  // Create json message body asset and push to `message`
}
```

To count for those cases we can write some tests for the following behaviours in `test/unit/parser/oas/parseMediaTypeObject-test.js`:

- we don't generate a message body for application/xml content-type
- we don't generate a message body for application/json WHEN the user defined `example` inside the `Media Type Object`, the example takes presidence. There is a test "creates an messageBody asset from an example for JSON type" which can be used for reference.
- we generate an example in the case we have a simple structure that does not contain any references:
    
    ```yaml
    type: object
    example:
      name: doe
    ```

- I'd write a test for a referenced dataStructure to ensure that the parser does not crash, you can borrow code from the existing test "parses a schema reference into as a data structure".

The file `test/unit/parser/oas/parseMediaTypeObject-test.js` should contain many tests that you can look at to get a grasp of how to write tests.

I hope my feedback here can help @alexylew, if you have any other questions do let me know. I'm happy to help you as much as I can, we could even jump on a call/pair program and walk through the changes together if it will help.

alexylew:
Thanks for the super detailed reply Kyle, really appreciate the effort :) This PR is very much just a rough stab at it, happy to trim down the scope and merge it in multiple pieces. I'm very much working on this with the goal of plugging in my very idiosyncratic OpenAPI 3.0 spec and getting Dredd tests out of it, so a lot of it is geared toward the way our spec is written. Specifically, we do a lot of examples on base level components and then `$ref`'ing them in other places. Because of that, this code tries to do the dereferencing in order to build up the top level examples without me having to go and add a bunch of top level examples. I do agree that we can tackle that in a later PR.

It sounds like the goal of this PR should be trimmed down to just adding support for the `examples` field on `MediaTypeObjects`, does that sound right?

kylef:
@alexylew ~~Ah, I though you was trying to solve the schema.example path not examples. Ok so I've created https://github.com/apiaryio/api-elements.js/pull/187 which adds the base infrastructure to parse the "Example Object". What we need to do is think about how a Media Type Object with multiple "Example Object" will be represented in API Elements. I believe we cannot map them 1-1, API Elements is built around the concept of "request-response" pairs (or transactions). If you have multiple request "examples", what does the responses look like? Perhaps we somehow match examples from response "examples".~~

Ok so I got carried away while trying to support you by providing the base infrastructure to do examples parsing. I ended up implementing a primitive version of examples in https://github.com/apiaryio/api-elements.js/pull/187. This should support examples and references which you are looking for. Would you like to give it a go and perhaps we can identify what is missing and we can point you in the right direction to implement.

PR #187, opened 2019-03-14
Parse "Example Object" in OAS 3

kylef:
This PR adds support for parsing example object, at this point we don't support any items inside the example object. This PR is setting up the infrastructure to start parsing the examples. This component has been integrated into the components parser, but it has not been integrated into the media type object yet. We should think about how that even works and how examples maps into API Elements to be able to acheive that.


kylef:
Scratch my previous comment, I got a bit carried away and added base implementation where we support "examples" in the Media Type Object. We only support one example which has "value" (note, we will warn if the user supplied 2 or more examples, we need to figure out how to map multiple examples, I've just gone down the MVP examples route). Examples is only supported for application/json as we do that for 'example'. We can fix these together separately, I think it is simple to remove this restriction.

See https://github.com/OAI/OpenAPI-Specification/blob/39d133d34a0acae12ed66fc0a25f7ec4dc1b9391/versions/3.0.0.md#media-type-examples for examples of how examples is used.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
```diff
@@ -9,40 +9,40 @@ const {
...
-  };
-
-  const parseJSONExample = (example) => {
-    const body = JSON.stringify(example.value.toValue());
-    const asset = new namespace.elements.Asset(body);
-    asset.classes.push('messageBody');
-    asset.contentType = mediaType;
-    return asset;
-  };
+const isJSONMediaType = (mediaType) => {
```
honzajavorek: This one could be `function`

PR #188, opened 2019-03-15
Fury support detection by method

klokane:
Fury now internally call `detect()` method with parameter `method`
It allows to adapter distinguish among supported method for given
operation instead "catch all" implementation.

Calling with `method` parameter is optional old backward compatible
style is fully supported as before.

Support `mediaTypes` two diferent formats, first is "catch all" and
backward compatilble array:
```
AdapterWithArrayMediaTypes = {
 ...
  mediaTypes: [
    'text/vnd.parseable',
    'text/vnd.serializable',
  ]
  ...
}
```

And advanced definition which better granularity to allows distinguish
among individual supported mediaTypes with relation to operation

```
AdapterWithObjectMediaTypes = {
 ...
  mediaTypes: {
    parse: [ 'text/vnd.parseable' ],
    serialize: [ 'text/vnd.serializable' ],
  ...
}
```

### DIFF
--- a/packages/fury/README.md
+++ b/packages/fury/README.md
@@ -117,15 +117,41 @@ fury1.parse(...);
 
 Adapters convert from an input format such as API Blueprint into refract elements. This allows a single, consistent interface to be used to interact with multiple input API description formats. Writing your own adapter allows you to add support for new input formats.
 
+Note about `mediaTypes`: it allows two kinds of definitions. First one is **array** type. It is intended to "catch all" implementation.
+It is useful if your adapter implements only one of fury methods, or if all methods accepts same types of Media Type.
+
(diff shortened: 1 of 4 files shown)


packages/fury/README.md
```diff
@@ -117,15 +117,41 @@ fury1.parse(...);
...
+
+```js
+export const mediaTypes = ['text/vnd.my-adapter'];
+```
+
+If you need to distinguish among supported input Media Types for methods use:
+
+```js
+export const mediaTypes = {
+  parse: ['text/vnd.my-parsing', 'text/vnd.another-supported-parsing],
```
kylef:
```suggestion
  parse: ['text/vnd.my-parsing', 'text/vnd.another-supported-parsing'],
```

klokane:
@kylef sry, I rebase changes by hand into one patch and push force them

PR #190, opened 2019-03-15
Update Node requirements to >= 6

kylef:
We already dropped Node 4 when we removed it from CI. Some package.json information was outdated.


honzajavorek:
Shouldn't it bump version and get a mention in the changelog?

kylef:
@honzajavorek For fury-adapter-oas3-parser I don't think so, it's not a new requirement as fury-adapter-oas3-parser requires you to have Fury core which already requires Node >= 6.

As for swagger-zoo, that's a different game as that doesn't depend on the core. It doesn't have a changelog yet (I am wondering if it makes sense, the changes 99% of the time shadow fury-adapter-swagger as they are fixtures tied to specific parser version -- @honzajavorek opinions welcome here). It does deserve a major release of swagger-zoo, which is indicated in commit message (`BREAKING CHANGE:`), I realise now I should have broken this into two commits as only swagger-zoo is a breaking change. We don't always do the releases in same PR as the changes. At the moment it is depending on what the change is, usually bug fixes result in a release (https://github.com/apiaryio/api-elements.js/pull/189 for example).

For example currently there are changes in Fury core, and swagger-zoo which are unreleased:

```
$ npx lerna changed -l
fury-adapter-apiary-blueprint-parser
fury-adapter-apib-parser
fury-adapter-apib-serializer
fury-adapter-oas3-parser
fury-adapter-remote
fury-adapter-swagger
fury-cli
fury
swagger-zoo
lerna success found 9 packages ready to publish
```

We have some outstanding internal changes to Fury remaining (for fury-adapter-remote) which should be wrapped up today and then I was planning to release another beta release of the entire Fury toolchain.

honzajavorek:
I was concerned mainly about the adapter, but the reasoning makes sense. I think `swagger-zoo` is fine, at least on npm it has zero dependents, so I'd say: whatever works for us

PR #191, opened 2019-03-18
Support generating message body from schema

kylef:


packages/fury-adapter-oas3-parser/test/integration/fixtures/components/operation-object-requestBody.json
```diff
@@ -142,6 +142,27 @@
...
+                              ]
+                            }
+                          },
+                          "attributes": {
+                            "contentType": {
+                              "element": "string",
+                              "content": "application/json"
+                            }
+                          },
+                          "content": "{}"
```
kylef: If you're wondering why these are empty, the schema in these fixtures is `type: object` with no further information. As these fixtures are generally testing a specific behaviour (such as referencing) and not the various abilities of the schema parser).

PR #192, opened 2019-03-19
fix(core): method is propagated form fury to `adapter.dectect()`

klokane:


packages/fury/test/fury-test.js
```diff
@@ -312,13 +312,40 @@ describe('Fury class', () => {
...
+      before(() => {
+        fury = new Fury();
+        fury.use(detectingAdapter);
+      });
+
+      it('detect with supported method', () => {
+        assert.deepEqual(fury.detect('test', 'parse'), [detectingAdapter]);
+      });
+
+      it('do not detect without method', () => {
```
kylef:
```suggestion
      it('does not detect without method', () => {
```

packages/fury/test/fury-test.js
```diff
@@ -312,13 +312,40 @@ describe('Fury class', () => {
...
+        assert.deepEqual(fury.detect('test'), []);
+      });
+
+      it('do not detect with unsupported method', () => {
```
kylef:
```suggestion
      it('does not detect with unsupported method', () => {
```

PR #199, opened 2019-03-25
fix(remote): use mediatype from Fury

klokane:


packages/fury-adapter-remote/lib/adapter.js
```diff
@@ -46,18 +51,28 @@ class FuryRemoteAdapter {
...
+      });
+      return mediaTypes.includes(mediaType);
+    }
+    return false;
   }
 
-  parse({ source, minim }, cb) {
-    const inputMediaType = detectMediaType(source, this.options.defaultInputMediaType);
+  parse({ source, minim, mediaType }, cb) {
+    const inputMediaType = mediaType || detectMediaType(source, this.options.defaultInputMediaType);
```
kylef: `defaultInputMediatype` was removed in the rest of PR. Should it still be here?

klokane: You are right, I miss it

klokane:
Updated

PR #202, opened 2019-03-26
Release entire toolchain

kylef:
Please do not merge and leave the rest to me once approved.

 - api-elements@0.1.1
 - fury-adapter-apiary-blueprint-parser@3.0.0-beta.7
 - fury-adapter-apib-parser@0.14.0
 - fury-adapter-apib-serializer@0.10.0
 - fury-adapter-oas3-parser@0.7.0
 - fury-adapter-remote@0.1.0
 - fury-adapter-swagger@0.25.0
 - fury-cli@0.8.12
 - fury@3.0.0-beta.10
 - minim-api-description@0.9.1
 - minim-parse-result@0.11.1
 - swagger-zoo@3.0.0

I discovered some minor inconsistencies in changelogs which have also been addressed (such as incorrect date format, incorrect past versions).

### DIFF
--- a/packages/fury-adapter-apib-serializer/package.json
+++ b/packages/fury-adapter-apib-serializer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fury-adapter-apib-serializer",
-  "version": "0.9.1",
+  "version": "0.10.0",
   "description": "API Blueprint serializer for Fury.js",
   "author": "Apiary.io <support@apiary.io>",
   "license": "MIT",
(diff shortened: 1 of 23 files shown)


packages/fury-adapter-apib-serializer/package.json
```diff
@@ -20,12 +20,12 @@
...
   },
   "peerDependencies": {
-    "fury": "3.0.0-beta.9"
+    "fury": "3.0.0-beta.10"
   },
   "devDependencies": {
     "chai": "^4.1.2",
     "eslint": "^5.9.0",
-    "fury": "3.0.0-beta.9",
+    "fury": "3.0.0-beta.10",
```
opichals: Strict dev-dep, but the apib-parser uses `^`. Is it intentional?

kylef: I've updated fury-adapter-apib-parser and fury-adapter-swagger to use strict "3.0.0-beta.10" as we have everywhere else. Good catch, I missed that one. Fighting lerna here as it is trying to use `^` with major beta release which isn't safe as any future beta can introduce incompatibilities.

Although in our case it does not really matter because Fury is in same repository and that is used to fulfil the dependency which is developed along side all packages in this repo.

PR #206, opened 2019-04-01
Finished scheme support

pksunkara:
This PR finishes the full support for oauth2 flows (after the breaking change in the Oauth2 Scheme element)

**UPDATED**: No more breaking changes ([Spec PR](https://github.com/apiaryio/api-elements/pull/55))

### DIFF
--- a/packages/fury-adapter-oas3-parser/STATUS.md
+++ b/packages/fury-adapter-oas3-parser/STATUS.md
@@ -72,7 +72,7 @@ Key:
 | responses | [~](#responses-object) |
 | callbacks | [✕](https://github.com/apiaryio/api-elements.js/issues/74) |
 | deprecated | ✕ |
-| security | [✕](https://github.com/apiaryio/api-elements.js/issues/77) |
+| security | ✓ |
 | servers | [✕](https://github.com/apiaryio/api-elements.js/issues/76) |
 
 ## Parameter Object
--- a/packages/fury-adapter-oas3-parser/STATUS.md
+++ b/packages/fury-adapter-oas3-parser/STATUS.md
@@ -72,7 +72,7 @@ Key:
 | responses | [~](#responses-object) |
 | callbacks | [✕](https://github.com/apiaryio/api-elements.js/issues/74) |
 | deprecated | ✕ |
-| security | ✓ |
+| security |  [✕](https://github.com/apiaryio/api-elements.js/issues/329)  |
 | servers | [✕](https://github.com/apiaryio/api-elements.js/issues/76) |
 
 ## Parameter Object
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
@@ -158,7 +158,7 @@ function parseComponentsObject(context, element) {
         if (value) {
           if (value instanceof namespace.elements.AuthScheme) {
             // eslint-disable-next-line no-param-reassign
-            value.meta.id = key.clone();
+            value.id = key.clone();
             array.push(value);
 
             return;
@@ -167,7 +167,7 @@ function parseComponentsObject(context, element) {
           // append oauth2 flow names
           value.forEach((flow) => {
             // eslint-disable-next-line no-param-reassign
-            flow.meta.id = `${key.toValue()} ${flow.grantTypeValue}`;
+            flow.id = `${key.toValue()} ${flow.grantTypeValue}`;
             array.push(flow);
           });
         }
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
@@ -156,9 +156,20 @@ function parseComponentsObject(context, element) {
 
       object.forEach((value, key) => {
         if (value) {
-          // eslint-disable-next-line no-param-reassign
-          value.meta.id = key.clone();
-          array.push(value);
+          if (value instanceof namespace.elements.AuthScheme) {
+            // eslint-disable-next-line no-param-reassign
+            value.meta.id = key.clone();
+            array.push(value);
+
+            return;
+          }
+
+          // append oauth2 flow names
+          value.forEach((flow) => {
+            // eslint-disable-next-line no-param-reassign
+            flow.meta.id = `${key.toValue()} ${flow.grantTypeValue}`;
+            array.push(flow);
+          });
         }
       });
 
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOauthFlowsObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOauthFlowsObject.js
@@ -72,6 +72,10 @@ function parseOauthFlowsObject(context, object) {
       authScheme.push(new namespace.elements.Member('grantType', grantTypes[member.key.toValue()]));
       authScheme.push(member.value.getMember('scopes'));
 
+      R.filter(R.is(namespace.elements.Transition), member.value).forEach((item) => {
+        authScheme.push(item);
+      });
+
       return authScheme;
     }));
 
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
@@ -112,8 +112,12 @@ function parseOperationObject(context, path, member) {
         return requirement.get(0);
       }
 
-      const allOf = new namespace.elements.Array(requirement.content);
-      allOf.element = 'allOf';
+      const link = new namespace.elements.Link();
+      link.relation = 'profile';
+      link.href = 'https://github.com/refractproject/rfcs/issues/39';
+
+      const allOf = new namespace.elements.Extension(requirement.content);
+      allOf.meta.set('links', new namespace.elements.Array([link]));
 
       return allOf;
     }));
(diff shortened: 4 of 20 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseOauthFlowsObject.js
```diff
@@ -66,13 +66,18 @@ function parseOauthFlowsObject(context, object) {
     parseObject(context, name, parseMember),
     flows => new namespace.elements.Array(flows.content),
     R.map((member) => {
-      const authScheme = new namespace.elements.AuthScheme();
+      const authSchemeFlow = new namespace.elements.Array();
 
-      authScheme.element = 'Oauth2 Scheme';
```
pksunkara: This part will be part of another PR. Removed here temporarily.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSecuritySchemeObject.js
```diff
@@ -64,6 +59,16 @@ function validateHttpScheme(context, securityScheme) {
   return parseObject(context, name, parseMember, ['scheme'], [], true)(securityScheme);
 }
 
+function validateOauth2Scheme(context, securityScheme) {
+  const parseMember = R.cond([
+    [hasKey('flows'), parseOauthFlowsObject],
+
+    [R.T, e => e],
```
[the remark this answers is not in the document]
pksunkara: It's better to this in a separate PR because there are a lot of places this needs to be done.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSecurityRequirementObject.js
```diff
@@ -39,25 +39,21 @@ function parseSecurityRequirementObject(context, object) {
...
+      } else {
+        e = new namespace.elements.Object({});
+      }
 
-        if (scopes.length) {
-          e = new namespace.elements.Object({ scopes });
-        } else {
-          e = new namespace.elements.Object({});
-        }
+      e.element = member.key.toValue();
```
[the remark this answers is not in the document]
pksunkara: This file is not changed anymore, but the tests have been added

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSecuritySchemeObject.js
```diff
@@ -110,14 +115,16 @@ function parseSecuritySchemeObject(context, object) {
...
-    // R.when(isOauth2Scheme, parseSecuritySchemeFlowsObject),
+    R.when(isOauth2Scheme, R.curry(validateOauth2Scheme)(context)),
     (securityScheme) => {
       const authScheme = new namespace.elements.AuthScheme();
 
       const type = securityScheme.getValue('type');
       const scheme = securityScheme.getValue('scheme');
 
-      if (type === 'apiKey' || (type === 'http' && scheme === 'bearer')) {
+      if (type === 'oauth2') {
```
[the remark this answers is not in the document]
pksunkara: I am not sure I want to do that. It would crash our system, right?

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSecurityRequirementObject.js
```diff
@@ -40,16 +40,17 @@ function parseSecurityRequirementObject(context, object) {
   const parseSecurityRequirement = pipeParseResult(namespace,
     parseObject(context, name, parseMember),
     (securityRequirement) => {
+      // TODO: expand oauth requirements into multiples depending on flows
```
pksunkara: This is the only thing that's not implemented in the whole of security area. If the swagger specifies an oauth2 based security scheme, we are expanding it into multiple security schemes so that we don't break the APIE regarding flows. So, when the user uses that security scheme in a security requirement, we need to replace that security requirement with the multiple security schemes we created above.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
```diff
@@ -156,9 +156,20 @@ function parseComponentsObject(context, element) {
...
+            value.meta.id = key.clone();
+            array.push(value);
+
+            return;
+          }
+
+          // append oauth2 flow names
+          value.forEach((flow) => {
+            // eslint-disable-next-line no-param-reassign
+            flow.meta.id = `${key.toValue()} ${flow.grantType}`;
```
kylef: Context has a method `registerId()` which can be used to register IDs and know if they are unique. If there is a schema with the `id` which could clash we can figure it out and error out.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
```diff
@@ -102,13 +104,25 @@ function parseOperationObject(context, path, member) {
     ),
   ]));
 
+  const parseSecurity = pipeParseResult(namespace,
+    R.unless(isArray, createWarning(namespace, `'${name}' 'security' is not an array`)),
+    R.compose(R.chain(parseSecurityRequirementObject(context)), R.constructN(1, namespace.elements.Array)),
```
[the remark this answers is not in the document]
pksunkara: This isn't working. I think there's some misunderstanding here. There are 2 nested levels here and you are probably thinking there's only 1.

pksunkara:
> Context has a method registerId() which can be used to register IDs and know if they are unique. If there is a schema with the id which could clash we can figure it out and error out.

I want to do this in a separate PR. Also, security scheme (does it exist?) validation is not done since it's tied into the above. The splitting of oauth2 schemes into different flows when we encounter them in security requirements is also not done since it is also tied into the above 2 things that are pending.

packages/fury-adapter-oas3-parser/STATUS.md
```diff
@@ -72,7 +72,7 @@ Key:
 | responses | [~](#responses-object) |
 | callbacks | [✕](https://github.com/apiaryio/api-elements.js/issues/74) |
 | deprecated | ✕ |
-| security | [✕](https://github.com/apiaryio/api-elements.js/issues/77) |
+| security | ✓ |
```
kylef: security isn't fully supported, should be ~ with link to nessecery objects below which show the unsupported components:

- provides an unsupported warning for openIdConnect type
- provides warning for unsupported bearerFormat property
- provides warning for unsupported openIdConnectUrl property

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSecuritySchemeObject.js
```diff
@@ -110,20 +115,35 @@ function parseSecuritySchemeObject(context, object) {
...
       const scheme = securityScheme.getValue('scheme');
+      const description = securityScheme.get('description');
+
+      if (type === 'oauth2') {
+        const flows = securityScheme.get('flows');
+
+        if (description) {
+          flows.forEach((flow) => {
+            // eslint-disable-next-line no-param-reassign
+            flow.description = description;
```
kylef: We're re-using the description below thus we should copy here. Else running `.freeze()` on the resultant parse result which meets both code paths will cause exception (in Dredd and V3 docs which use freeze()).

```suggestion
            flow.description = description.clone();
```

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseSecurityRequirementObject-test.js
```diff
@@ -97,7 +97,7 @@ describe('Security Requirement Object', () => {
     expect(arr.get(0).element).to.equal('customOauth2');
     expect(arr.get(0).length).to.equal(1);
 
-    const scopes = arr.get(0).get('scopes');
+    const scopes = arr.get(0).get(0).value;
```
kylef: Comment isn't specific to this line but for file (which GitHub don't let you comment on).

There are no tests for the case that the scheme is not registered in components. Thus when the scheme is not known the code may produce an invalid element:

```yaml
unknown: []
```

```yaml
element: unknown
```

----

Secondly, what happens if I call my element scheme the same as a pre-existing element?


```yaml
array: []
null: []
annotation: []
```

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
```diff
@@ -102,13 +104,28 @@ function parseOperationObject(context, path, member) {
...
+    requirements => requirements.map((requirement) => {
+      if (requirement.length === 1) {
+        return requirement.get(0);
+      }
+
+      const allOf = new namespace.elements.Array(requirement.content);
+      allOf.element = 'allOf';
```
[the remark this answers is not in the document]
pksunkara: We decided not to do the extension and in fact implement the `allOf` element.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js
kylef: Moved to https://github.com/apiaryio/api-elements.js/issues/330

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseSecurityRequirementObject-test.js
pksunkara: The first part will be part of a different PR and the second part is already discussed above.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSecurityRequirementObject.js
pksunkara: Part of next PR

PR #208, opened 2019-04-02
Handle invalid security scheme components

kylef:
The parser is currently crashing for certain inputs, for example:

```yaml
openapi: 3.0.0
info:
  title: abc
  version: xyz
paths: {}
components:
  securitySchemes:
    Foo: {}
```

```
TypeError: Cannot read property 'meta' of undefined
    at object.forEach (/Users/kyle/Projects/apiaryio/api-elements.js/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js:154:17)
    at content.forEach.item (/Users/kyle/Projects/apiaryio/api-elements.js/node_modules/minim/lib/primitives/ObjectElement.js:203:63)
    at Array.forEach (<anonymous>)
    at ObjectElement.forEach (/Users/kyle/Projects/apiaryio/api-elements.js/node_modules/minim/lib/primitives/ObjectElement.js:203:25)
    at pipeParseResult (/Users/kyle/Projects/apiaryio/api-elements.js/packages/fury-adapter-oas3-parser/lib/parser/oas/parseComponentsObject.js:152:14)
    at run (/Users/kyle/Projects/apiaryio/api-elements.js/packages/fury-adapter-oas3-parser/lib/pipeParseResult.js:60:25)
    at XWrap.f (/Users/kyle/Projects/apiaryio/api-elements.js/node_modules/ramda/src/reduceWhile.js:40:27)
    at XWrap.@@transducer/step (/Users/kyle/Projects/apiaryio/api-elements.js/node_modules/ramda/src/internal/_xwrap.js:12:17)
    at _arrayReduce (/Users/kyle/Projects/apiaryio/api-elements.js/node_modules/ramda/src/internal/_reduce.js:11:34)
    at _reduce (/Users/kyle/Projects/apiaryio/api-elements.js/node_modules/ramda/src/internal/_reduce.js:45:12)
```

### DIFF
--- a/packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseComponentsObject-test.js
+++ b/packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseComponentsObject-test.js
@@ -322,6 +322,24 @@ describe('Components Object', () => {
       expect(securitySchemes.get(0)).to.be.instanceof(namespace.elements.AuthScheme);
       expect(securitySchemes.get(0).meta.id.toValue()).to.equal('token');
     });
+
+    it('handles invalid security scheme', () => {
+      const components = new namespace.elements.Object({
+        securitySchemes: {
+          Basic: null,
+        },
+      });
+
+      const parseResult = parse(context, components);
+      expect(parseResult.length).to.equal(2);
+
+      const parsedComponents = parseResult.get(0);
+      expect(parsedComponents).to.be.instanceof(namespace.elements.Object);
+
+      const schemes = parsedComponents.get('securitySchemes');
+      expect(schemes).to.be.instanceof(namespace.elements.Array);
+      expect(schemes.isEmpty).to.be.true;
+    });
   });
 
   describe('#examples', () => {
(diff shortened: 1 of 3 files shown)


packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseComponentsObject-test.js
```diff
@@ -322,6 +322,24 @@ describe('Components Object', () => {
...
+      });
+
+      const parseResult = parse(context, components);
+      expect(parseResult.length).to.equal(2);
+
+      const parsedComponents = parseResult.get(0);
+      expect(parsedComponents).to.be.instanceof(namespace.elements.Object);
+
+      const schemes = parsedComponents.get('securitySchemes');
+      expect(schemes).to.be.instanceof(namespace.elements.Array);
```
kylef: @pksunkara I don't think we should return an array here but instead an object like the other components parsers. This is inconsistent with all other component parsers and means we cannot use generic functions like `parseReference` as it relies on an object being here.

Outside scope of this fix though but will be needed to handle dereferencing these values.

pksunkara: I think the fix should be part of this PR or it should wait until both this and the main security PR is merged which might take some time. IMHO, should be part of this PR.

kylef: You mentioned holding off on merging this to prevent a complicated merge/rebase for #206. I don't think that will be a problem, #206 does not touch parseComponentsObject or related tests.

PR #210, opened 2019-04-03
Improve YAML error message descriptions

kylef:
The YAML parser sometimes includes "problem" in the error which contains a clearer message on the underlying problem. When available we can include this in our error annotations. For example if the parser hits a tab in the document which is not valid, it was showing an error "while scanning for the next token", we can be more detailed and include the problem so the error message can become: "found character \t that cannot start any token, while scanning for the next token".

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/parseYAML.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/parseYAML.js
@@ -106,8 +106,17 @@ function parse(source, context) {
   try {
     ast = yaml.compose(source);
   } catch (error) {
+    let message;
+
+    if (error.problem) {
+      const problem = error.problem.replace('\t', '\\t');
+      message = `${problem}, ${error.context}`;
+    } else {
+      message = error.context;
+    }
+
     const annotation = new namespace.elements.Annotation(
-      `YAML Syntax: ${error.context}`,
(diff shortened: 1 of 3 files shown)


packages/fury-adapter-oas3-parser/lib/parser/parseYAML.js
```diff
@@ -106,8 +106,17 @@ function parse(source, context) {
   try {
     ast = yaml.compose(source);
   } catch (error) {
+    let message;
+
+    if (error.problem) {
+      const problem = error.problem.replace('\t', '\\t');
```
kylef: Since a tab isn't clear in the error message (it would just show as whitespace), I am replacing tabs specifically the literal value '\t'. The error message would read as:

> error: YAML Syntax: found character \t that cannot start any token, while scanning for the next token - line 1452

artem-zakharchenko: Should we be aware of other symbols in the error messages, or is it just tabs?

kylef: Perhaps, tab is the only one I'm aware of so far. Other types of errors from the YAML parser which would cause similiar problems haven't come up yet.

Maybe there is a generic way we can make a representation of a string to expose all unreadable characters.

PR #213, opened 2019-04-09
feat(oas3): validate parameter name according to location

klokane:


packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseParameterObject-test.js
```diff
@@ -62,6 +61,78 @@ describe('Parameter Object', () => {
...
+        in: 'header',
+      });
+
+      const parseResult = parse(context, parameter);
+
+      expect(parseResult.length).to.equal(1);
+      expect(parseResult).to.contain.warning('\'Parameter Object\' \'name\' in location \'header\' ignore keywords \'Accept\', \'Content-Type\' and \'Authorization\'');
+    });
+
+    it('does not allow header name to contain reserved keyword `Content-Type`', () => {
```
kylef: We should have one more case of this type of test for `Authorization`.
```suggestion
    it('does not allow header name to contain reserved keyword `Authorization`', () => {
      const parameter = new namespace.elements.Object({
        name: 'Authorization',
        in: 'header',
      });

      const parseResult = parse(context, parameter);

      expect(parseResult.length).to.equal(1);
      expect(parseResult).to.contain.warning('\'Parameter Object\' \'name\' in location \'header\' ignore keywords \'Accept\', \'Content-Type\' and \'Authorization\'');
    });
 
    it('does not allow header name to contain reserved keyword `Content-Type`', () => {
```

packages/fury-adapter-oas3-parser/test/integration/fixtures/petstore.sourcemap.json
```diff
@@ -251,31 +251,6 @@
                       "content": {
                         "key": {
                           "element": "string",
-                          "attributes": {
```
kylef: This is a regression, source maps for the name are being removed.

Above we have the line:

```js
parameter.set('name', encodeQueryName(parameter.get('name').toValue()));
```

This should become something like the following so we can retain the sourcemaps in the "name" element:

```js
const name = parameter.get('name');
name.content = encodeQueryName(name.toValue());
```

packages/fury-adapter-oas3-parser/lib/parser/oas/parseParameterObject.js
```diff
@@ -106,11 +106,33 @@ function parseParameterObject(context, object) {
     [R.T, createInvalidMemberWarning(namespace, name)],
   ]);
 
-  const isPathParameter = parameter => parameter.getValue('in') === 'path';
+  const createUnsupportedNameError = createError(namespace, `'${name}' 'name' contains unsupported characters. Only alphanumeric characters are currently supported`);
+
+  const createReservedHeaderNamesWarning = createWarning(namespace, `'${name}' 'name' in location 'header' ignore keywords 'Accept', 'Content-Type' and 'Authorization'`);
```
kylef: I think we could make this warning message a bit clearer:

```suggestion
  const createReservedHeaderNamesWarning = createWarning(namespace, `'${name}' 'name' in location 'header' should not be 'Accept', 'Content-Type' or 'Authorization'`);
```

Issue #217, opened 2019-04-12
Swagger: unhandled `throw` from function `lookupReference`


klokane:
If there is non existing reference in swagger document `lookupReference` will throw exception
https://github.com/apiaryio/api-elements.js/blob/master/packages/fury-adapter-swagger/lib/json-schema.js#L70

It is not handled on upper level, we should handle it by transformation to error/warning in parse result

kylef:
Here's the stacktrace:

```json
{
    "errorMessage": "Schema reference must be reference to #/definitions",
    "errorType": "Error",
    "stackTrace": [
        "lookupReference (/var/task/node_modules/fury-adapter-swagger/lib/json-schema.js:50:11)",
        "dereference (/var/task/node_modules/fury-adapter-swagger/lib/json-schema.js:108:17)",
        "example.map.value (/var/task/node_modules/fury-adapter-swagger/lib/json-schema.js:115:33)",
        "Array.map (<anonymous>)",
        "dereference (/var/task/node_modules/fury-adapter-swagger/lib/json-schema.js:115:20)",
        "_.forEach (/var/task/node_modules/fury-adapter-swagger/lib/json-schema.js:122:21)",
        "/var/task/node_modules/lodash/lodash.js:4944:15",
        "baseForOwn (/var/task/node_modules/lodash/lodash.js:3001:24)",
        "/var/task/node_modules/lodash/lodash.js:4913:18",
        "Function.forEach (/var/task/node_modules/lodash/lodash.js:9359:14)"
    ]
}
```

kylef:
In theory, the following should be able to reproduce, given the stacktrace is from dereferenceing the examples when example is an array:

```yaml
swagger: '2.0'
info:
  title: My API
  version: 1.0.0
produces:
  - application/json
paths:
  /:
    get:
      responses:
        200:
          description: foo
          schema:
            type: object
            example:
              - $ref: '#/info'
```

However it parses and gives an error message to user instead of an uncaught error:

> error: (4) Schema reference must be reference to #/definitions - line 13

kylef:
Managed to reprouce, the structure must be found inside definitions:

```yaml
swagger: '2.0'
info:
  title: My API
  version: 1.0.0
paths: {}
definitions:
  Test:
    type: object
    example:
      - $ref: '#/info'
```

PR #221, opened 2019-04-24
Handle invalid schema object examples in OAS 2

kylef:
Fixes https://github.com/apiaryio/api-elements.js/issues/217

More context can be found in the large comment I added inside the function that is altered below.

### DIFF
--- a/packages/fury-adapter-swagger/lib/json-schema.js
+++ b/packages/fury-adapter-swagger/lib/json-schema.js
@@ -6,6 +6,8 @@ if (!Object.values) {
   values.shim();
 }
 
+class ReferenceError extends Error {}
+
 // Test whether a key is a special Swagger extension.
 const isExtension = (value, key) => _.startsWith(key, 'x-');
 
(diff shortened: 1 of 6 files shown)


packages/fury-adapter-swagger/lib/json-schema.js
```diff
@@ -93,6 +95,20 @@ const pathHasCircularReference = (paths, path, reference) => {
...
+  // See https://github.com/apiaryio/api-elements.js/issues/220
+  //
+  // At thsi point, changing that behaviour would be a significant breaking
+  // change and it will affect some of our larger users. Not to mention that
+  // swagger-parser will still dereference the examples in cases where our code
+  // path doesn't, it won't be easy to solve.
+  //
+  // The below code attemps to dereference an example, but if we can't we
+  // will just return the example (possibly a "reference object") to be
+  // the example value.
```
kylef: I want to point out, the information in this comment is not new. I think it is worth clarifying what the function does and why. In the past there are also cases where invalid reference would be treated as example value, this PR just adds one additional case.

packages/fury-adapter-swagger/test/fixtures/schema-references-example-invalid.yaml
```diff
@@ -0,0 +1,10 @@
+swagger: '2.0'
+info:
+  title: My API
+  version: 1.0.0
+paths: {}
+definitions:
+  Test:
```
kylef: This 'Schema Object' declares an array type which has an sample value of an array. The array's first value is an object which has a key `$ref` and the value `#/info`, it is not a reference as per the spec and we are not treating it as one since it doesn't reference any valid component in the document.

Issue #37, opened 2018-12-11
Fury: Provide async/await API


kylef:
We should update `fury` to have an async (promise returning) API. This has been started in Protagonist and will depend on the release of Drafter 4.

```js
const { Fury } = require('fury');
const fury = new Fury();
const apiBlueprintParser = require('fury-adapter-apib-parser');
fury.use(apiBlueprintParser);
const openAPI2Parser = require('fury-adapter-swagger');
fury.use(openAPI2Parser);

const source = `
FORMAT: 1A

# My API
## GET /message
+ Response 200 (text/plain)

        Hello World!
`;

const parseResult = await fury.parse({source});
```

- [x] Support async/await in Fury core for validate (`packages/fury`)
- [x] Can use "callbackify"/"promisify" to simplify the process as first step. The logic can be contained within Fury to wrap existing adapters methods in `promisify`
- [x] Simplify some tests to use async/await
- [x] Update all examples in each package README to use `async/await` syntax.
- [x] Add CHANGELOG entry into `packages/fury` in "Enhancements"

char0n:
The only disadvantage here is that we'll not be able to distinguish expected failures and unexpected errors. But with promises, basically every API suffers from this problems and we got used to it.

opichals:
> we'll not be able to distinguish expected failures and unexpected errors. But with promises, basically every API suffers from this problems and we got used to it.

@kylef Perhaps we could introduce a convention for the _callback errors_ to be thrown wrapped into a special Error class to be able to distinguish from others?

char0n:
@opichals there is now standard [error.code](https://nodejs.org/api/errors.html#errors_error_code) property which you can use. The recommended approach is to create new Error type and set  the `code` property internally. Then instead of inspecting the `type` with `instanceof`  of the Error, inspecting it's shape by using `code` property is recommended.

kylef:
Available in Fury 3.0.0-beta.11 (https://github.com/apiaryio/api-elements.js/releases/tag/fury%403.0.0-beta.11)

PR #222, opened 2019-04-26
Make the adapter interface use promises

kylef:
As a part of #37, I've made the interfaces between Fury core and the adapters use promises. This is a relatively simple change, the biggest changes are in the tests.

I've made this a draft PR as I would like to release https://github.com/apiaryio/api-elements.js/pull/221 beforehand. I think we can review this and I can add release commits as we will need to make an entire toolchain release.

### DIFF
--- a/packages/fury-adapter-apiary-blueprint-parser/lib/adapter.js
+++ b/packages/fury-adapter-apiary-blueprint-parser/lib/adapter.js
@@ -9,10 +9,10 @@ const mediaTypes = [
 
 const detect = source => mediaTypes.indexOf(deckardcain.identify(source)) !== -1;
 
-const parse = (options, done) => {
+function parse(options) {
   const parser = new Parser(options);
-  parser.parse(done);
-};
+  return Promise.resolve(parser.parse());
+}
 
 module.exports = {
   name, mediaTypes, detect, parse,
--- a/packages/fury-adapter-apib-serializer/test/adapter-test.js
+++ b/packages/fury-adapter-apib-serializer/test/adapter-test.js
@@ -6,13 +6,16 @@
 
 const { expect } = require('chai');
 const fs = require('fs');
-const fury = require('fury');
+const { Fury } = require('fury');
 const glob = require('glob');
 const path = require('path');
-const { serialize } = require('../lib/adapter');
+const adapter = require('../lib/adapter');
 const { indent } = require('../lib/filters');
 
 const base = path.join(__dirname, 'fixtures');
+
+const fury = new Fury();
+fury.use(adapter);
 
 describe('API Blueprint serializer adapter', () => {
   const files = glob.sync(path.join(base, '*.json'));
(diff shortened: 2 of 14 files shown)


packages/fury-adapter-apib-serializer/test/adapter-test.js
```diff
@@ -35,7 +38,7 @@ describe('API Blueprint serializer adapter', () => {
         return done(loadErr);
       }
 
-      return serialize({ api }, (serializeErr, serialized) => {
+      return fury.serialize({ api }, (serializeErr, serialized) => {
```
kylef: In this test we was testing the serializers adapter interface, I think it's better to test against Fury interface so we know the integration completely works.

packages/fury/test/fury-test.js
```diff
@@ -409,22 +409,24 @@ describe('Fury class', () => {
   describe('pass mediaType into adapter operations', () => {
     const fury = new Fury();
 
-    function test(first) {
-      return function impl(second) { assert.equal(first, second); };
+    function test(expectedMediaType) {
```
kylef: This test took me a while to get my head around it so I tried to re-name some of the parts to make it simpler. This test also was throwing a string which is tricky because we can't reject a promise with a string (it shows a warning when you do that), it's probably not a good idea to do anyway.

I've updated so we return a parse result containing the media type instead.

packages/fury-adapter-apiary-blueprint-parser/lib/adapter.js
```diff
@@ -9,10 +9,10 @@ const mediaTypes = [
 
 const detect = source => mediaTypes.indexOf(deckardcain.identify(source)) !== -1;
 
-const parse = (options, done) => {
+function parse(options) {
   const parser = new Parser(options);
-  parser.parse(done);
-};
+  return Promise.resolve(parser.parse());
```
artem-zakharchenko: Maybe we should also reject in case `parser.parse()` throws an exception? So we could use `.catch()` on `parse` invocations:

```js
parse(options).catch(console.error)
```

packages/fury/README.md
```diff
@@ -155,28 +155,28 @@ export function detect(source[, method]) {
   return source.match(/some-test/i) !== null;
 }
 
-export function parse({minim, generateSourceMap, mediaType, source}, done) {
+export async function parse({minim, generateSourceMap, mediaType, source}) {
```
artem-zakharchenko: This comment may be irrelevant, as I don't know if `async` is used anywhere else, but in case this is the only usage, I'd recommend to use `Promise` instead. `async/await` appends generators to support this feature in the compiled code, while in this case it's only in order for `parse`, `validate` (and others) to return a Promise.

Returning a plain Promise may reduce the package's size.

kylef: We're using Promises everywhere because we still support Node 6, however the documentation here uses async for brevity/simplicity. I think developers are smart enough to use Promises if they use older Node.

I don't see how it would change the package size, the Promise syntax if anything will be a little longer and thus make the README file sizes larger?

PR #228, opened 2019-05-16
Clone parameter to prevent duplicate instance in tree

kylef:
We should not place the exact same instance of a a referenced element in a final parse result tree multiple times, otherwise `.freeze()` won't work. Final ParseResult should contain unique instances in all cases and thus we must clone an element in these cases.

Fixes https://github.com/apiaryio/dredd/issues/1360

### DIFF
--- a/packages/fury-adapter-oas3-parser/test/integration/components-test.js
+++ b/packages/fury-adapter-oas3-parser/test/integration/components-test.js
@@ -21,14 +21,21 @@ describe('components', () => {
     });
   });
 
-  it("'Media Type Object' schema references", () => {
-    const file = path.join(fixtures, 'media-type-object-schema');
-    return testParseFixture(file);
-  });
+  describe('Media Type Object', () => {
+    it('handles schema references', () => {
+      const file = path.join(fixtures, 'media-type-object-schema');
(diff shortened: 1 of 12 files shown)


packages/fury-adapter-oas3-parser/test/integration/components-test.js
```diff
@@ -4,14 +4,21 @@ const testParseFixture = require('./testParseFixture');
...
-  it("'Path Item Object' parameter referencing unsupported parameter", () => {
-    const file = path.join(fixtures, 'path-item-object-parameters-unsupported-parameter');
-    return testParseFixture(file);
+  describe('Path Item Object', () => {
+    it('handles parameter references', () => {
+      const file = path.join(fixtures, 'path-item-object-parameters');
+      return testParseFixture(file);
+    });
+
+    it('handles multiple references to same parameter', () => {
```
kylef: Using an integration test because this particular bug is with interplay of using parameters referenced from multiple different components with element `freeze()`. There's a lot of setup state for this and it's simpler to test referencing in this way.

kylef:
I've also added similiar tests to ensure that all similiar referencing behaves correctly too.

PR #229, opened 2019-05-16
Fix source map locations for OAS 3 warnings

kylef:
I discovered missing source maps on two types of warnings, will comment below.

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
@@ -96,7 +96,7 @@ function parseMediaTypeObject(context, MessageBodyClass, element) {
       }
 
       if (examples.length > 1) {
-        parseResult.push(createWarning(namespace, `'${name}' 'examples' only one example is supported, other examples have been ignored`, examples));
+        parseResult.push(createWarning(namespace, `'${name}' 'examples' only one example is supported, other examples have been ignored`, examples.content[1]));
       }
 
       return parseResult;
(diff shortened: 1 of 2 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
```diff
@@ -96,7 +96,7 @@ function parseMediaTypeObject(context, MessageBodyClass, element) {
       }
 
       if (examples.length > 1) {
-        parseResult.push(createWarning(namespace, `'${name}' 'examples' only one example is supported, other examples have been ignored`, examples));
+        parseResult.push(createWarning(namespace, `'${name}' 'examples' only one example is supported, other examples have been ignored`, examples.content[1]));
```
kylef: "examples" is an array and thus doesn't have source maps, we should point to the 1st example which causes the warning for source maps.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -131,8 +131,13 @@ function validateValuesMatchSchema(context, schema) {
 function parseSchema(context) {
   const { namespace } = context;
 
-  const ensureValidType = R.unless(isValidType, createWarning(namespace,
-    `'Schema Object' 'type' must be either ${types.join(', ')}`));
+  const ensureValidType = R.unless(
+    isValidType,
+    R.compose(
+      createWarning(namespace, `'Schema Object' 'type' must be either ${types.join(', ')}`),
+      getValue
```
kylef: "type" the element passed ot createWarning is a member element and not a value and thus does not contain source maps.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
honzajavorek: Isn't this an ideal case for providing multiple source map ranges? 😉 Not that I need them, this solution is fine for Dredd.

PR #233, opened 2019-05-29
Prevent dynamic creation of element classes

kylef:
This one is gonna be tricky to review as it touches a lot of the code, but I never really changed it. Most of the diff is because the indentation is significantly changed as classes have been defined inside a function which is no longer the case. The diff will treat these changes as entire additions (as file was renamed to match class name as per our style guide).

Now we will no longer create classes dynamically inside a function which improves the simpliciy of the project and it can make implementation such as in `valueOf` simpler:

```js
  // TODO e instanceof EnumElement fails because of dependency injection
  // checking for e.element === 'enum' as a temporary  walkaround
  const isEnumElement = e => (e.element === 'enum' || e instanceof EnumElement);
```

^ I'll fix this after as I can add additional tests as it will fix a bug (subclasses of referenced Enum not handled correctly).


kylef:
> Not really an expert on how minim got into supporting direct inheritance from ES6 classes instead of functions

Supported that for a long time, uptown uses the prototype chain and thus ES6 inheritance worked. These days Minim is in ES6 using classes too though.

Issue #236, opened 2019-02-01
JSON Body generation of array of object doesn't respect object example


kylef:
Likely a limitation of https://github.com/json-schema-faker/json-schema-faker.

Take the following example:

```yaml
swagger: '2.0'
info:
  version: '1.0'
  title: "constantin"
produces:
- application/json
paths:
  /questions:
    get:
      summary: List All Questions
      responses:
        200:
          description: Successful Response
          schema:
            type: array
            items:
              $ref: '#/definitions/Question'
definitions:
  Question:
    title: Question
    type: object
    properties:
      question:
        type: string
        example: hi
      published_at:
        type: string
        example: 2019
      choices:
        type: array
        items:
          $ref: '#/definitions/Choice'
    required:
      - question
      - published_at
      - choices
  Choice:
    title: Choice
    type: object
    properties:
      votes:
        type: integer
        format: int32
        example: 512
      choice:
        type: string
        example: Swift
    required:
      - votes
      - choice
```

Produces the following JSON Body:

```json
[
  {
    "question": "hi",
    "published_at": "2019",
    "choices": []
  }
]
```

Whereas users would expect:

```json
[
  {
    "question": "hi",
    "published_at": "2019",
    "choices": [
      {
        "votes": 512,
        "choice": "Swift"
      }
    ]
  }
]
```

PR #237, opened 2019-06-04
Add regression fixture for json body with array[object]

kylef:
This PR fixes the issue described at https://github.com/apiaryio/api-elements.js/issues/236. The underlying problem was in JSON Schema Faker and updating to the newer version actually solve this problem. I've added the original test fixture in a seaprate commit (https://github.com/apiaryio/api-elements.js/commit/3cecd2cbdde39e5a6184b71d6189be08095329a6) so that the diff for (https://github.com/apiaryio/api-elements.js/commit/be64f25c1b85814f201a3511b05c01db728babc6) shows you the changes to the produced JSON object.

Updating to json-schema-faker rc17 brings some other behaviour changes so I've had to provide a constant random function of 0 otherwise the amount of items in the array is random and it makes it harder to test. This breaks other tests and test fixtures randomly. Using constant random also helps us provide determanistic results so I think it's a better option.

### DIFF
--- a/packages/fury-adapter-swagger/test/generator-test.js
+++ b/packages/fury-adapter-swagger/test/generator-test.js
@@ -53,7 +53,7 @@ describe('bodyFromSchema', () => {
     const asset = bodyFromSchema(schema, payload, parser, 'application/json');
 
     expect(typeof asset.content).to.equal('string');
-    expect(asset.content.length).to.be.above(0);
+    expect(asset.content.length).to.equal(0);
   });
 
   it('limits a strings min/max length to 256', () => {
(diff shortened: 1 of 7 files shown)


packages/fury-adapter-swagger/test/generator-test.js
```diff
@@ -53,7 +53,7 @@ describe('bodyFromSchema', () => {
     const asset = bodyFromSchema(schema, payload, parser, 'application/json');
 
     expect(typeof asset.content).to.equal('string');
-    expect(asset.content.length).to.be.above(0);
+    expect(asset.content.length).to.equal(0);
```
kylef: This is now constantly an empty string because the change to randomness, this now matches the behaviour of API Blueprint where an element of type string without any example, default values would emit an empty string.

PR #240, opened 2019-06-05
Ensure test expectations cause failure

kylef:
Each test case was not calling `done` afterwards and thus the test
failure could happen outside mocha as the API is async. The test was not waiting for the operation to complete. This is exactly the case with the last test which
was failing as the expectation was incorrect but mocha returns 0 as all the tests technically passed.

After pushing to CI I experianced a failure due to one of the tests causing a timeout, since these tests are hitting HTTP API there can be a delay and thus caused the timeout to happen. So I've double the timeout interval.

### DIFF
--- a/packages/fury-adapter-remote/test/fury-test.js
+++ b/packages/fury-adapter-remote/test/fury-test.js
@@ -7,7 +7,9 @@ const swaggerSource = '{ "swagger": "2.0", "info": { "version": "1.0.0", "title"
 const invalidSwaggerSource = '{ "swagger": "2.0", "info": { "version": 1, "title": "swg" } }';
 
 
-describe('Adapter works with Fury interface (with default config)', () => {
+describe('Adapter works with Fury interface (with default config)', function remoteAdapterTests() {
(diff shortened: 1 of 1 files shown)


packages/fury-adapter-remote/test/fury-test.js
```diff
@@ -258,11 +275,12 @@ describe('Adapter works with Fury interface', () => {
       }));
     });
 
-    it('parses blueprint with mediaType', () => {
+    it('parses blueprint with mediaType', (done) => {
       fury.parse({ source: blueprintSource, mediaType: 'text/vnd.apiblueprint' }, (err, result) => {
         expect(result).to.be.undefined;
         expect(err).to.be.instanceof(Error);
-        expect(err).to.have.property('message', 'Request failed with status code 415');
+        expect(err).to.have.property('message', 'Request failed with status code 404');
```
kylef: This is in a test `with invalid endpoint` (one level up in describe) and thus the API returns 404 for hitting an unknown path.

Issue #28, opened 2018-12-04
Rename furys `.minim` property to `.namespace`


kylef:
I don't think we should be interchanging the terms `minim` and `namespace`. Minim refers to the library "minim" whereas namespace refers to the instance of the minim `Namespace` type.

Right now we are calling the namespace instance `minim` in the interface of the `Fury` package.

PR #241, opened 2019-06-05
Rename minim option as namespace

kylef:
Part of https://github.com/apiaryio/api-elements.js/issues/28, the `minim` option passed between Fury and the adapters has now been renamed to `namespace` which is clearer. "minim" is the library and "namespace" is the instance of Namespace from minim.

BREAKING CHANGE: The interface between Fury and the adapters now provides the Minim Namespace under the key `namespace` (previously `minim`)


pksunkara:
Do you want to add the commits to release versions here itself? Only fury would be breaking and it would be an rc version anyway, right?

kylef:
I'll make the releases separate, will try make 1-2 more PRs first with some other refactoring.

Lerna will force us to release next fury RC before due to interlinked packages so no risk of breaking other releases.

PR #242, opened 2019-06-06
Test on Node 12

kylef:
Node 12 has reached current version, let's make sure it works for users to aid future upgrading (https://nodejs.org/en/about/releases/).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,17 +15,22 @@ defaults: &defaults
 
 jobs:
   test-node6:
+    <<: *defaults
     docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -15,18 +15,23 @@ defaults: &defaults
 
 jobs:
   test-node6:
+    <<: *defaults
```
kylef: I've also moved this from below the `docker` line because it doesn't seem as clear that "docker" is overriding the inherited defaults (which contain docker).

PR #243, opened 2019-06-06
Enable Node 12 tests on CI

kylef:
Missing change from https://github.com/apiaryio/api-elements.js/pull/242. There is one remote test which fails because they've changed the error message in Node JS, before the error message was:

> getaddrinfo ENOTFOUND some.stupid.non.existing.domain some.stupid.non.existing.domain:80

Now it is:

> ENOTFOUND some.stupid.non.existing.domain

I've used regex matching to support both, you can see from CI builds.

### DIFF
--- a/packages/fury-adapter-remote/test/fury-test.js
+++ b/packages/fury-adapter-remote/test/fury-test.js
@@ -256,7 +256,7 @@ describe('Adapter works with Fury interface', () => {
       fury.parse({ source: blueprintSource, mediaType: 'text/vnd.apiblueprint' }, (err, result) => {
         expect(result).to.be.undefined;
         expect(err).to.be.instanceof(Error);
-        expect(err).to.have.property('message', 'getaddrinfo ENOTFOUND some.stupid.non.existing.domain some.stupid.non.existing.domain:80');
+        expect(err.message).to.match(/ENOTFOUND some\.stupid\.non\.existing\.domain/);
         done();
       });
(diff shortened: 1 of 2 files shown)


packages/fury-adapter-remote/test/fury-test.js
```diff
@@ -256,7 +256,7 @@ describe('Adapter works with Fury interface', () => {
       fury.parse({ source: blueprintSource, mediaType: 'text/vnd.apiblueprint' }, (err, result) => {
         expect(result).to.be.undefined;
         expect(err).to.be.instanceof(Error);
-        expect(err).to.have.property('message', 'getaddrinfo ENOTFOUND some.stupid.non.existing.domain some.stupid.non.existing.domain:80');
+        expect(err.message).to.match(/ENOTFOUND some.stupid.non.existing.domain/);
```
honzajavorek: shouldn't the dots be escaped?

kylef: I guess so, I'll do that. `.` in regex is anything so it works

honzajavorek: I know, I approved because I don't think this changes much.

kylef: Yeah, I've addressed in 5bad13bd7a7eacb593df4408a59f1086b6a38569

PR #245, opened 2019-06-06
Support async/await (and promises)

kylef:
Implements support for async/await (and promises) under https://github.com/apiaryio/api-elements.js/issues/37.

Effectively, the following work:

```js
const parseResult = await fury.parse({ source });
console.log(parseResult.api.title.toValue());
```

```js
fury
  .parse({ source })
  .then((parseResult) => {
    console.log(parseResult.api.title.toValue());
  });
```

```js
// existing behaviour with callbacks
fury.parse({ source }, (err, parseResult) => {
  console.log(parseResult.api.title.toValue());
});
```

~This depends on https://github.com/apiaryio/api-elements.js/pull/244 as I am using async/await syntax in the tests which won't work on Node 6.~

### DIFF
--- a/packages/fury/test/parse-test.js
+++ b/packages/fury/test/parse-test.js
@@ -1,10 +1,11 @@
 const { assert } = require('chai');
 const { Fury } = require('../lib/fury');
+const { rejects } = require('./assert');
 
 describe('Parser', () => {
   let fury;
 
-  before(() => {
+  beforeEach(() => {
     fury = new Fury();
     const adapter = {
       name: 'passthrough',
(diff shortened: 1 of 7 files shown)


packages/fury/test/parse-test.js
```diff
@@ -17,64 +18,99 @@ describe('Parser', () => {
     fury.use(adapter);
   });
 
-  it('should parse through mediatype', (done) => {
-    fury.parse({ source: 'dummy', mediaType: 'text/vnd.passthrough' }, (err, result) => {
-      assert.equal(result.content, 'dummy');
-      done(err);
+  describe('using callback', () => {
```
kylef: Diff noise, I've only indented all these tests to be inside the 'using callback' describe.

PR #247, opened 2019-06-07
Add hrefVariables property to HttpRequest

kylef:
During https://github.com/apiaryio/Paw-APIBlueprintImporter/pull/35 I realised that `hrefVariables` property was missing on HttpRequest element so this PR brings it (the property exists on other elements which can have the hrefVariables attribute).

I've also noticed when I was looking for the relevant tests that none of the properties in this file was tested, so my first commit was adding this missing test coverage.

### DIFF
--- a/packages/api-elements/CHANGELOG.md
+++ b/packages/api-elements/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Support for NodeJS 6 has been removed, upgrading to NodeJS 8 or newer is
   recommended.
 
+### Enhancements
+
+- Added the `hrefVariables` property to `HttpRequest` element.
+
 ## 0.1.1 (2019-03-26)
 
 ### Enhancements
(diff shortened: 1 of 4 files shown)


kylef:
@opichals / @honzajavorek just wanted to let you know, I assume you are using these properties. This PR adds `hrefVariables` property to `HttpRequest`. This will make it into the next release of `api-elements`.

honzajavorek:
Something tells me I worked around the missing property in the past, so perhaps some code in DT can be simplified.

kylef:
@honzajavorek Yes, (from dredd-transactions):

```
compile/compileURI/index.js:  // not have the '.hrefVariables' convenience property yet. If it's added in
compile/compileURI/index.js:  // the future, '.attributes.get('hrefVariables')' can be replaced
compile/compileURI/index.js:  // with '.hrefVariables'.
compile/compileURI/index.js:    .map(element => compileParams(element.attributes.get('hrefVariables')))
```

PR #248, opened 2019-06-07
feat(oas3): support parameters in media types

### LINKED ISSUE #246: Add support for parameters in media types for OAS 3
Reported from https://github.com/apiaryio/dredd/issues/1391, in our OpenAPI 3 parser, media types that do not have parameters for example `application/json; chatset=utf-8` are not permitted. This is because we use media-typer module which does not permit them, instead we should use the content module. The logic is in a function `isValidMediaType` in `parseMediaTypeObject` https://github.com/apiaryio/api-elements.js/blob/b6bb8a1b5fb6177d78247d1eb7d3a122bd2de647/packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js#L49-L56

We should swap out media-typer with content-type (https://www.npmjs.com/package/content-type) to solve.

I've already made a failing test to demonstrate the problem, we should make this test pass:

```diff
diff --git a/packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseMediaTypeObject-test.js b/packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseMediaTypeObject-test.js
index 07f2e46..799d38a 100644
--- a/packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseMediaTypeObject-test.js
+++ b/packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseMediaTypeObject-test.js
@@ -29,6 +29,16 @@ describe('Media Type Object', () => {
     expect(parseResult).to.contain.warning("'Media Type Object' media type 'foo' is invalid");
   });

+  it('permits media type with parameters', () => {
+    const mediaType = new namespace.elements.Member('application/json; charset=utf-8', {});
+
+    const parseResult = parse(context, messageBodyClass, mediaType);
+
+    const message = parseResult.get(0);
+    expect(message).to.be.instanceof(messageBodyClass);
+    expect(message.contentType.toValue()).to.equal('application/json; chatset=utf-8');
+  });
+
   it('returns a HTTP message body', () => {
     const mediaType = new namespace.elements.Member('application/json', {});
```

klokane:
closes #246

Solved by replacing `media-typer` by `content-type` as sugested in #246

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
@@ -1,5 +1,5 @@
 const R = require('ramda');
-const mediaTyper = require('media-typer');
+const contentType = require('content-type');
 const pipeParseResult = require('../../pipeParseResult');
 const {
   isExtension, hasKey, getKey, getValue,
@@ -19,11 +19,9 @@ const unsupportedKeys = ['encoding'];
 const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
 
 function isJSONMediaType(mediaType) {
-  const type = mediaTyper.parse(mediaType);
-  return (
-    type.type === 'application'
-    && (type.subtype === 'json' || type.suffix === 'json')
-  );
+  const type = contentType.parse(mediaType);
+  const jsonMediaType = /^application\/\S*json$/i;
+  return jsonMediaType.test(type.type);
 }
 
 const createJSONMessageBodyAsset = R.curry((namespace, mediaType, value) => {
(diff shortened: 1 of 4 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
```diff
@@ -19,11 +19,9 @@ const unsupportedKeys = ['encoding'];
 const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
 
 function isJSONMediaType(mediaType) {
-  const type = mediaTyper.parse(mediaType);
-  return (
-    type.type === 'application'
-    && (type.subtype === 'json' || type.suffix === 'json')
-  );
+  const type = contentType.parse(mediaType);
+  const jsonMediaType = /^application\/\S*json$/i;
```
kylef: I wonder if it would make sense to also use mediaTyper in this case, just wondering if the regex may be fragile in some cases, but I cannot think of any. So up to you really.

```js
const type = mediaTyper.parse(mediaType);
const type = mediaTyper = parse(type.type);
return (
    type.type === 'application'
    && (type.subtype === 'json' || type.suffix === 'json')
)
```

PR #249, opened 2019-06-10
Treat 'Example Object' with length property as object

kylef:
I discovered the following abort while parsing an OAS 2 document:

```
5: serialiseContent [/app/node_modules/minim/lib...
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
1: node::Abort()
```

It turns out that if you have `length: 500000` or similiar in an Example Object, the parser would end up treating the example as an array while creating a minim ObjectElement from the array, this results in an object with 500,000 member elements for key 0 through 500,000. When serialising this as JSON you may hit out of memory errors.

For example, if you had a schema such as:

```yaml
definitions:
  User:
    type: object
    example:
      length: 500000
```

The underlying cause is a use of Lodash `forEach` function while iterating over the values during parsing the document. The lodash `forEach` method treats objects with a length property as an array. For example:

```js
> require('lodash').forEach({ length: 20 }, (value, key) => { console.log(key) })
0
1
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
```

### DIFF
--- a/packages/fury-adapter-swagger/CHANGELOG.md
+++ b/packages/fury-adapter-swagger/CHANGELOG.md
@@ -27,6 +27,23 @@
 
   [#236](https://github.com/apiaryio/api-elements.js/issues/236)
 
+- While parsing an 'Example Object' (found in a 'Schema Object') which contains
+  an object with a property `length` anywhere in the example tree. The example
+  object will be interpreted as an array of the given length. If the value of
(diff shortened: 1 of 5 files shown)


kylef:
I've audited the entire repository's use of `_.forEach` and I identified a couple more in the OAS 2 parser. Added commit af10f1b to change these to `forIn`.

PR #255, opened 2019-06-19
Performance optimisation for valueOf

kylef:
This PR brings a performance optimisation for `valueOf`, there was a use of `filter()[0]` which can be replaced for `find()` which prevents looping over the entire collection unnessecerily. This brings a decent performance improvement when you are running `valueOf` a bunch of times while parsing a very large OpenAPI 3 document which is generating data structures for hundreads of data structures. Seeing around a 7-10 seconds improvement for the particular document I have been testing.

Subsequently I've made minor changes to remove unnessecery work (see below)

### DIFF
--- a/packages/api-elements/lib/define-value-of.js
+++ b/packages/api-elements/lib/define-value-of.js
@@ -22,7 +22,7 @@ function hasTypeAttribute(e, attribute) {
   if (undefined !== e._attributes) {
     const attrs = e.attributes.get('typeAttributes');
     if (undefined !== attrs && undefined !== attrs.content) {
-      return undefined !== attrs.content.find(attr => attr.content === attribute);
+      return attrs.content.some(attr => attr.content === attribute);
     }
   }
   return false;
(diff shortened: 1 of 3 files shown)


packages/api-elements/lib/define-value-of.js
```diff
@@ -93,10 +93,8 @@ module.exports = (namespace) => {
     return undefined;
   }
 
-  // TODO e instanceof EnumElement fails because of dependency injection
-  // checking for e.element === 'enum' as a temporary  walkaround
-  const isEnumElement = e => (e.element === 'enum' || e instanceof EnumElement);
```
kylef: The problem being worked around here was solved in a prior PR, you can see the tests continue to pass after this change.

packages/api-elements/lib/define-value-of.js
```diff
@@ -93,10 +93,8 @@ module.exports = (namespace) => {
...
-  const isPlural = e => (e instanceof ArrayElement) || (e instanceof ObjectElement);
```
kylef: ObjectElement is an instanceof ArrayElement and thus this additional check was redundant.

PR #261, opened 2019-07-01
Simplify valueOf by removing dynamic nature of functions created in namespace

kylef:
This PR is broken into separate commits, I recommend reading the respective commits and commit messages for further information.

I've removed the passing down the namespace to `defineValueOf` because it is unnessecery and can cause multiple versions of some of the functions used to exist at runtime depending on how many times you create a namespace.


kylef:
Added 4e047cb for consistency

PR #272, opened 2019-07-02
Setup a smoke test

kylef:
This PR adds a smoke test to API Elements, which installs Fury, all of the Fury adapters and then the fury-cli. We then run a few operations on Fury such as parsinrg a document of all description formats (to test all of the adapters). I've also ran `--version`, `--help` to verify they return 0. Note `--version` outputs versions of all configured adapters.

Check out the full output of https://circleci.com/gh/apiaryio/api-elements.js/2834

This did actually help me catch a bug that `npm pack` doesn't build part of fury-adapter-apiary-blueprint-parser (fixed in 7b4dcaa).

### DIFF
--- a/.circleci/smoke.sh
+++ b/.circleci/smoke.sh
@@ -12,7 +12,7 @@ create_package() {
 install() {
   PACKAGE="$1"
 
-  TARBALL="$(cd packages/$PACKAGE && npm pack)"
+  TARBALL="$(cd packages/$PACKAGE && npm pack | tail -n1)"
   mv "packages/$PACKAGE/$TARBALL" "$TMPDIR"
 
   cd "$TMPDIR" && npm install "$TARBALL" && cd "$API_ELEMENTS"
--- a/.circleci/smoke.sh
+++ b/.circleci/smoke.sh
@@ -34,7 +34,7 @@ npx fury --version
 
 # Parse Apiary Blueprint
 cat << EOF | npx fury -
-HOST: http://www.google.com/
+HOST: http://example.com/
 
 --- Sample API v2 ---
 ---
--- a/.circleci/smoke.sh
+++ b/.circleci/smoke.sh
@@ -2,11 +2,11 @@
 
 set -e
 
-TMPDIR="$(mktemp -d 2>/dev/null || mktemp -d -t 'apiejs-smoketest')"
-API_ELEMENTS="$(pwd)"
+TMPDIR="$(mktemp -d)"
+PROJECT_DIR="$(pwd)"
 
 create_package() {
-  cd "$TMPDIR" && npm init --yes && cd "$API_ELEMENTS"
+  cd "$TMPDIR" && npm init --yes && cd "$PROJECT_DIR"
 }
 
 install() {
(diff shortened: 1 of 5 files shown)


.circleci/smoke.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+
+set -e
+
+TMPDIR="$(mktemp -d 2>/dev/null || mktemp -d -t 'apiejs-smoketest')"
```
kylef: Just to mention, this is to count for platform differences (BSD mktemp vs GNU mktemp). *one* of these will paths will pass to create the tmp directory.

honzajavorek: That's maybe something which would be worth also as an inline comment right above that line.

.circleci/smoke.sh
```diff
@@ -0,0 +1,72 @@
...
+API_ELEMENTS="$(pwd)"
```
honzajavorek: I'd rather call this something like `PROJECT_DIR` so it's clearer what's in that variable, but YMMV

.circleci/smoke.sh
```diff
@@ -0,0 +1,72 @@
...
+install fury-adapter-apib-serializer
+install fury-cli
+
+cd "$TMPDIR"
+npx fury --help
+npx fury --version
+
+# Parse Apiary Blueprint
+cat << EOF | npx fury -
+HOST: http://www.google.com/
```
honzajavorek:
```suggestion
HOST: http://www.example.com/
```

.circleci/smoke.sh
```diff
@@ -0,0 +1,72 @@
...
+
+--- Sample API v2 ---
+---
+Welcome to the our sample API documentation. All comments can be written in (support [Markdown](http://daringfireball.net/projects/markdown/syntax) syntax)
+---
+EOF
+
+# Parse API Blueprint
+cat << EOF | npx fury -
```
honzajavorek: Do you want to assert somehow that the output is valid JSON/YAML as opposed to e.g. an exception? Even though... exception is non-zero exit code, so the script would halt - okay then.

.circleci/smoke.sh
kylef: Ah. I didn't even look at the contets of this file much (copied it from the test fixture from the subpackage).

Fixed in 9bab9f9 (I've also updated test fixture I copied it from).

.circleci/smoke.sh
kylef: Changed in 9889e3f

.circleci/smoke.sh
kylef: Yes, if there is any exception then Node would exit 1.

It would be neat to be able to assert that the result is valid JSON. Perhaps even that it is an object which has `element` of `parseResult` etc. I don't think it is that trivial to add though, I don't think it's possible to do it in a one-liner with node. In Python we could do `python -c 'import json, sys; print(json.load(sys.stdin))'` (this is Node image without Python).

honzajavorek: You redirect output to file, then cat the file (for a reference), and then something like

```bash
node -e 'JSON.parse(require("fs").readFileSync("foo"))'
```

honzajavorek: Forgot encoding, needs either toString() or readFileSync("foo", "utf8")

PR #273, opened 2019-07-03
Work around infinite loop bug when using pattern and minLength in OAS 2

kylef:
The changelog entry in diff explains the problem that I am working around. Effectively, if you have a pattern `^[A-z]*$` and a `minLength` of `1` then the pattern with `*` can resolve to an empty string which doesn't match validation. I'm working around it by detecting this particular issue and "patching" the regex to use `+` instead of `*` so that it is a "one or more" not "zero or more".

This is due to the current behaviour with JSON Schema Faker which gets into an infinite loop in this state. I think it is seeing that the pattern string doesn't generate the correct length and then it tries again, and again (forever). As we've seeded our "randomness" to produce consistent results it gets stuck.

This is a short-sighted solution, there are likely other cases this can happen which I am not counting for. This particular fix is to solve a majority problem that I am seeing a small amount of our customers hit (and subsequent platform instability). A proper fix should be made in json-schema-faker, that's more time consuming (and involves waiting for a fix to be released which given JSF release cycles can be months or more). So I've made this interim fix to solve platform instablity and customers not being able to use their OAS 2 document in our products. I've created APIARY-5932 to track this going forward (and for us to solve the problem in JSON Schema Faker).

I did consider providing a warning annotation to the user, but given how this part of the code base is architected, it wouldn't be possible to have source maps for the annotation without larger refactoring (which may in turn slow the parser down).

### DIFF
--- a/packages/fury-adapter-swagger/lib/json-schema.js
+++ b/packages/fury-adapter-swagger/lib/json-schema.js
@@ -166,6 +166,15 @@ const convertSubSchema = (schema, references, swagger) => {
     actualSchema.type = 'string';
   }
 
+  if (schema.pattern && schema.minLength && schema.pattern.startsWith('^[') && schema.pattern.endsWith(']*$')) {
+    // If a schema has a minimal length (minLength) > 0 AND there is a regex
+    // such as: `^[A-z]*$`, the schema can resolve to an empty string which
+    // doesn't match minLength.
+    //
+    // JSON Schema Faker will fail in that case and get into an infinite loop.
+    actualSchema.pattern = schema.pattern.replace('*$', '+$');
+  }
+
   if (schema.example) {
     actualSchema.examples = [dereference(schema.example, swagger)];
   }
--- a/packages/fury-adapter-swagger/CHANGELOG.md
+++ b/packages/fury-adapter-swagger/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Fury Swagger Parser Changelog
 
-## Master
+## 0.27.1 (2019-06-03)
 
 ### Bug Fixes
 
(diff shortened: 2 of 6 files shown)


packages/fury-adapter-swagger/lib/json-schema.js
```diff
@@ -166,6 +166,15 @@ const convertSubSchema = (schema, references, swagger) => {
     actualSchema.type = 'string';
   }
 
+  if (schema.pattern && schema.minLength && schema.pattern.startsWith('^[') && schema.pattern.endsWith(']*$')) {
+    // If a schema has a minimal length (minLength) > 0 AND there is a regex
+    // such as: `^[A-z]*$`, the schema can resolve to an empty string which
+    // doesn't match minLength.
+    //
+    // JSON Schema Faker will fail in that case and get into an infinite loop.
+    actualSchema.pattern = schema.pattern.replace('*$', '+$');
```
tjanc: Is this replacing all occurrences of `*$`? From what I understand from this PR we should only modify the end.

tjanc: Is there a way to make the modification closer to the invocation of that faker tool?

kylef: `$` refers to the end of the line in regex. So replacing `*$` (usually would happen once at the end), if it is in the schema multiple times I think it should be fine to replace it.

kylef: Unfortuantely not, without copying every schema and traversing over it at the JSON Schema Faker entrypoint. I would like to avoid that, it would be a far more expensive operation.

PR #300, opened 2019-07-19
Support generating text message body for string schemas

kylef:
This PR adds support for generating message body assets for text types like text/plain and text/html.

Example usage:

```yaml
responses:
  200:
    content:
      text/plain:
        schema:
          type: string
          example: The request was successful
```

To produce message body, such as if the response was:

```
Content-Type: text/plain

The request was successful
```

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
@@ -1,5 +1,6 @@
 const R = require('ramda');
-const contentType = require('content-type');
+const contentTyper = require('content-type');
+const mediaTyper = require('media-typer');
 const pipeParseResult = require('../../pipeParseResult');
 const {
   isExtension, hasKey, getKey, getValue,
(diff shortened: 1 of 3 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
```diff
@@ -19,9 +20,49 @@ const unsupportedKeys = ['encoding'];
 const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
 
 function isJSONMediaType(mediaType) {
-  const type = contentType.parse(mediaType);
-  const jsonMediaType = /^application\/\S*json$/i;
-  return jsonMediaType.test(type.type);
+  const contentType = contentTyper.parse(mediaType);
+  const { type, suffix, subtype } = mediaTyper.parse(contentType.type);
```
kylef: I've refactored isJSONMediaType to use mediaTyper so it is implemented in the same way that `isTextMediaType` is implemented. I think it can be a bit safer to use mediaTyper than custom regex.

PR #321, opened 2019-08-06
Support text-based examples in OAS 3

kylef:
Adds support for example values for text based media types such as `text/*` and `application/xml`. 'example' for XML is represented as a string value representing XML in OAS 3.0.

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
@@ -31,6 +31,12 @@ function isTextMediaType(mediaType) {
   return type === 'text';
 }
 
+function isXMLMediaType(mediaType) {
+  const contentType = contentTyper.parse(mediaType);
+  const { type, suffix, subtype } = mediaTyper.parse(contentType.type);
+  return type === 'application' && (suffix === 'xml' || subtype === 'xml');
+}
+
 const canGenerateMessageBodyForMediaType = R.anyPass([isJSONMediaType, isTextMediaType]);
 
 function generateMessageBody(context, mediaType, dataStructure) {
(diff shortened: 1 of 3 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
```diff
@@ -76,13 +82,27 @@ const createJSONMessageBodyAsset = R.curry((namespace, mediaType, value) => {
...
+
 function parseExample(namespace, mediaType) {
   const createExampleNotJSONWarning = createWarning(namespace,
-    "'Media Type Object' 'example' is only supported for JSON media types");
+    `'${name}' 'example' is not supported for media type '${mediaType}'`);
+
+  const isTextBasedType = R.anyPass([
+    () => isTextMediaType(mediaType),
+    () => isXMLMediaType(mediaType),
+  ]);
```
pksunkara: Shouldn't you be using R.curry?

packages/fury-adapter-oas3-parser/lib/parser/oas/parseMediaTypeObject.js
```diff
@@ -76,13 +82,27 @@ const createJSONMessageBodyAsset = R.curry((namespace, mediaType, value) => {
...
 
-  return R.ifElse(() => isJSONMediaType(mediaType),
-    R.compose(createJSONMessageBodyAsset(namespace, mediaType), getValue),
-    createExampleNotJSONWarning);
+  return R.compose(R.cond([
+    [() => isJSONMediaType(mediaType), createJSONMessageBodyAsset(namespace, mediaType)],
+    [() => isTextBasedType(mediaType), createTextMessageBodyAsset(namespace, mediaType)],
```
pksunkara: Why is the extra closure needed? Simple `isJSONMediaType(...)` should be enough.

kylef: It's the same as it was before `return R.ifElse(() => isJSONMediaType(mediaType),`. Otherwise the provided value to isJSONMediaType will be the example element and not the media type.

pksunkara: So, you are saying the following line should be changed:

https://github.com/apiaryio/api-elements.js/blob/master/packages/fury-adapter-oas3-parser/lib/parser/oas/parseParameterObject.js#L94

```
[() => hasKey('name'), parseName],
```

pksunkara: I am just wondering why the 2 cases are different. As far as I understand, they are not.

kylef: `[() => hasKey('name'), parseName],` <- hasKey here takes two arguments, it is a curried function taking a string (key name) and a member element, it returns boolean. This proposed change wouldn't work and cause tests to fail.

isTextBasedType takes a single argument, the media type name.

pksunkara: Yup, that was what I was missing in this context. Thanks.

PR #325, opened 2019-08-08
Release Updated Adapters

### LINKED ISSUE #314: non required fields should be converted to "optional"
Hi,
I have this field definition and i have explicitly set "required" to false:
```
{
    "type": "string",
    "description": "x",
     "name": "x",
      "in": "query",
       "required": false
}

```

This will generate:

```
+ Parameters
.....
   
    + sidecar
```

According to the spec, URI fields can be either `optional` or `required` and the default is required.

https://github.com/apiaryio/api-blueprint/blob/master/API%20Blueprint%20Specification.md#uri-parameters-section

I think this should apply for swagger fields whose "required" field is not set.

thanks!

kylef:
- fury-adapter-oas3-parser@0.9.1
- fury-adapter-swagger@0.27.2
    - swagger-zoo@swagger-zoo

Fixes #314

### DIFF
--- a/.circleci/smoke.sh
+++ b/.circleci/smoke.sh
@@ -23,6 +23,7 @@ echo "Setting up Fury in $TMPDIR"
 create_package
 install fury
 install fury-adapter-apib-parser
+install fury-adapter-swagger
 install fury-adapter-oas3-parser
 install fury-adapter-apiary-blueprint-parser
 install fury-adapter-apib-serializer
--- a/packages/fury-adapter-swagger/package.json
+++ b/packages/fury-adapter-swagger/package.json
@@ -1,11 +1,13 @@
 {
   "name": "fury-adapter-swagger",
-  "version": "0.27.1",
+  "version": "0.27.2",
   "description": "Swagger 2.0 parser for Fury.js",
   "author": "Apiary.io <support@apiary.io>",
   "license": "MIT",
   "main": "./lib/adapter.js",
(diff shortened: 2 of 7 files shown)


.circleci/smoke.sh
```diff
@@ -23,6 +23,7 @@ echo "Setting up Fury in $TMPDIR"
 create_package
 install fury
 install fury-adapter-apib-parser
+install fury-adapter-swagger
```
kylef: fury-adapter-swagger was not included in smoke tests and thus failed CI: https://circleci.com/gh/apiaryio/api-elements.js/3712?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

That's because it tried to install from NPM (and its unreleased).

packages/fury-adapter-swagger/package.json
```diff
@@ -1,11 +1,13 @@
 {
   "name": "fury-adapter-swagger",
-  "version": "0.27.1",
+  "version": "0.27.2",
   "description": "Swagger 2.0 parser for Fury.js",
   "author": "Apiary.io <support@apiary.io>",
   "license": "MIT",
   "main": "./lib/adapter.js",
-  "files": ["lib/*.js"],
+  "files": [
```
kylef: NPM/YARN reformatted file, will stick with the format the tools use so we don't have to fight it every time we do a release.

PR #403, opened 2020-01-29
Adds explicit "$schema" property to JSON Schema generated from Swagger

### LINKED ISSUE #399: Generated JSON Schemas do not contain $schema for draft-4
The openapi2 parser does not append `$schema` to generated JSON Schemas so tooling such as Gavel/Dredd do not interpret the JSON Schema as Draft 4 but instead Draft 3. The JSON Schema the parser generates is Draft 4 and thus incompatible.

The code here: https://github.com/apiaryio/api-elements.js/blob/6bd2df2e8cbbf7f966fcccd0844b41b0e50df40f/packages/fury-adapter-swagger/lib/json-schema.js#L343 generates JSON Schema. There is option copyDefinitions which indicates if it is generating the “root schema” instead of the sub schemas. Perhaps we should add a check there for copyDefinitions and then add $schema for draft-4 in here. Tests such as in https://github.com/apiaryio/api-elements.js/blob/6bd2df2e8cbbf7f966fcccd0844b41b0e50df40f/packages/fury-adapter-swagger/test/json-schema-test.js#L8 will need updating.

artem-zakharchenko:
## Changes

- Adds explicit `$schema` property in JSON Schema generated by OA2 document
- Adjusts `json-schema-test.js` to assert the presence of the root-level `$schema` property

## GitHub

- Fixes #399

### DIFF
--- a/packages/fury-adapter-swagger/CHANGELOG.md
+++ b/packages/fury-adapter-swagger/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Fury Swagger Parser Changelog
 
+## 0.28.1 (2020-01-30)
+
+### Bug Fixes
+
+- Sets an explicit `$schema` property on JSON Schema generated from a Swagger
+  document in `convertSchema`. Sets a JSON Schema Draft V4 as the value of the
+  `$schema` property, making generated schema implement a JSON Schema Draft V4.
+
 ## 0.28.0 (2019-12-06)
 
 ### Enhancements
--- a/packages/fury-adapter-swagger/lib/json-schema.js
+++ b/packages/fury-adapter-swagger/lib/json-schema.js
@@ -345,6 +345,8 @@ const convertSchema = (schema, root, swagger, copyDefinitions = true) => {
   const result = convertSubSchema(schema, references, swagger);
 
   if (copyDefinitions) {
+    result.$schema = 'http://json-schema.org/draft-04/schema#';
+
     if (references.length !== 0) {
       result.definitions = {};
     }
--- a/packages/fury-adapter-swagger/test/fixtures/auth-multi-consumes.json
+++ b/packages/fury-adapter-swagger/test/fixtures/auth-multi-consumes.json
@@ -169,7 +169,7 @@
                               "content": "application/schema+json"
                             }
                           },
-                          "content": "{\"type\":\"object\"}"
+                          "content": "{\"type\":\"object\",\"$schema\":\"http://json-schema.org/draft-04/schema#\"}"
                         },
                         {
                           "element": "dataStructure",
@@ -273,7 +273,7 @@
                               "content": "application/schema+json"
                             }
                           },
-                          "content": "{\"type\":\"object\"}"
+                          "content": "{\"type\":\"object\",\"$schema\":\"http://json-schema.org/draft-04/schema#\"}"
                         },
                         {
                           "element": "dataStructure",
--- a/packages/fury-adapter-swagger/test/fixtures/consumes-invalid-type.json
+++ b/packages/fury-adapter-swagger/test/fixtures/consumes-invalid-type.json
@@ -68,7 +68,7 @@
                               "content": "application/schema+json"
                             }
                           },
-                          "content": "{\"type\":\"object\"}"
+                          "content": "{\"type\":\"object\",\"$schema\":\"http://json-schema.org/draft-04/schema#\"}"
                         },
                         {
                           "element": "dataStructure",
--- a/packages/fury-adapter-swagger/test/fixtures/data-structure-generation-nullable-member.json
+++ b/packages/fury-adapter-swagger/test/fixtures/data-structure-generation-nullable-member.json
@@ -91,7 +91,7 @@
                               "content": "application/schema+json"
                             }
                           },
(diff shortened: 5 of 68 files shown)


artem-zakharchenko:
I'm currently experiencing failing tests under "Inherit Parameters" test. More details below.

During "Body Parameter" tests there is an assertion between the parsing result () and the actual data (taken from the source modified by `source.push(makeParameter())` before the relevant test):

https://github.com/apiaryio/api-elements.js/blob/3bbc63800fa41de7626453ff49bfd6aef2fdcf57/packages/fury-adapter-swagger/test/inherit-parameters-test.js#L235

The parsing result is valid, as it contains the added `$schema` property. The expected result from the source, however, doesn't contain that property, which makes assertion to fail. I'm looking at how the parameters are added via `makeParameter`:

https://github.com/apiaryio/api-elements.js/blob/3bbc63800fa41de7626453ff49bfd6aef2fdcf57/packages/fury-adapter-swagger/test/inherit-parameters-test.js#L46

This line is invoked in the first "Body Parameter" test when `makeParameter('test', 'body')` is pushed to the `source`. When I attempt to add the `$schema` property for two payloads to match I get a list of `Annotation` from Fury, indicating parsing errors:

```
[ Annotation {
    _content: 'Data does not match any schemas from \'oneOf\'',
    _storedElement: 'annotation',
    _meta:
     ObjectElement { _content: [Array], _storedElement: 'object' },
    _attributes:
     ObjectElement { _content: [Array], _storedElement: 'object' } },
  Annotation {
    _content: 'Source maps are only available with string input',
    _storedElement: 'annotation',
    _meta:
     ObjectElement { _content: [Array], _storedElement: 'object' },
    _attributes:
     ObjectElement { _content: [Array], _storedElement: 'object' } },
  Annotation {
    _content: 'Data does not match any schemas from \'oneOf\'',
    _storedElement: 'annotation',
    _meta:
     ObjectElement { _content: [Array], _storedElement: 'object' },
    _attributes:
     ObjectElement { _content: [Array], _storedElement: 'object' } },
  Annotation {
    _content: 'Missing required property: $ref',
    _storedElement: 'annotation',
    _meta:
     ObjectElement { _content: [Array], _storedElement: 'object' },
    _attributes:
     ObjectElement { _content: [Array], _storedElement: 'object' } },
  Annotation {
    _content: 'Additional properties not allowed: $schema',
    _storedElement: 'annotation',
    _meta:
     ObjectElement { _content: [Array], _storedElement: 'object' },
    _attributes:
     ObjectElement { _content: [Array], _storedElement: 'object' } },
  Annotation {
    _content: 'Missing required property: type',
    _storedElement: 'annotation',
    _meta:
     ObjectElement { _content: [Array], _storedElement: 'object' },
    _attributes:
     ObjectElement { _content: [Array], _storedElement: 'object' } } ]
```

From that list you can see that it complains about "Additional properties not allowed: $schema", but it also behaves differently as compared to when it's run without the added `$schema` property in `makeParameter`. 

I suppose adding that `$schema` property influences how the parsing is done? @kylef @pksunkara could you please advise me on this? Thank you.

kylef:
@artem-zakharchenko you will need to add the entry into the schema generated at https://github.com/apiaryio/api-elements.js/blob/8cd312061148abad63610a3b75afbe9944df32a6/packages/fury-adapter-swagger/test/inherit-parameters-test.js#L231. Note that "parameter.schema" is NOT JSON Schema, it's Swagger Schema. It just happens to be that they were the same in this case; but that no longer applies.

```js
const parameterSchema = source.paths['/'].parameters[0].schema; // but actually make copy of it so we don't modify it
parameterSchema.$schema = '...';
const schema = JSON.stringify(parameterSchema);
```

artem-zakharchenko:
@kylef any objection to isolate that logic into a helper utility?

```js
const schema = getSchemaFromSource(source)

function getSchemaFromSource(source) {
  const parameterSchema = _.clone(source)
  parameterSchema.$schema = '...'
  return JSON.stringify(oarameterSchema)
}
```

We're going to use this quite often in this test.

kylef:
The code in particular is deriving a JSON Schema from the Parameter's Swagger Schema and in doing so it will add the `$schema` now.

---

Could you add a part to `packages/fury-adapter-swagger/CHANGELOG.md` to specify the changes too, we have changelogs in this repo for each package:

```
## TBD

### Bug Fixes

- <some description of the problem we are solving>
```

artem-zakharchenko:
Should I squash updated fixtures commit into the `fix` one? Not sure how you approach this in the repo.

kylef:
> Should I squash updated fixtures commit into the fix one? Not sure how you approach this in the repo.

Yeah; although we can do that during merge (squash and merge in GH). I'd prefer to keep every commit passing tests so that tools like [`git bisect`](https://git-scm.com/docs/git-bisect) can be used. It makes it simpler if we need to do things like revert changes etc too.

kylef:
Changes look good :+1:, perhaps just adding changelog entry and we can merge this.

artem-zakharchenko:
> I'd prefer to keep every commit passing tests so that tools like git bisect can be used.

A valid point. I've squashed the commits and updated the relevant CHANGELOG. Please take a look. Thanks.

PR #411, opened 2020-02-13
feat(oas3): add `termsOfService` support

marcofriso:
Closes APIARY-6288

### DIFF
--- a/packages/fury-adapter-oas3-parser/CHANGELOG.md
+++ b/packages/fury-adapter-oas3-parser/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Fury OAS3 Parser Changelog
 
+## 0.10.2 (2020-02-17)
+
+### Enhancements
+
+- Adds support for the `info.termsOfService` schema extension.
+
 ## 0.10.1 (2020-01-30)
 
 ### Bug Fixes
--- a/packages/fury-adapter-oas3-parser/CHANGELOG.md
+++ b/packages/fury-adapter-oas3-parser/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Enhancements
 
-- Adds support for the `info.termsOfService` schema extension.
+- Adds support for `termsOfService` in 'Info Object'.
 
 ## 0.10.1 (2020-01-30)
 
--- a/packages/fury-adapter-oas3-parser/CHANGELOG.md
+++ b/packages/fury-adapter-oas3-parser/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Fury OAS3 Parser Changelog
 
-## 0.10.2 (2020-02-17)
+## Master
 
 ### Enhancements
 
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseInfoObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseInfoObject.js
@@ -35,17 +35,20 @@ const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
 function parseInfo(context, info) {
   const { namespace } = context;
 
-  const parseTermsOfService = (terms) => {
-    const link = new namespace.elements.Link();
-    link.relation = 'terms-of-service';
-    link.href = terms;
+  const parseTermsOfService = pipeParseResult(namespace,
+    parseString(context, name, false),
+    getValue,
+    (termsOfService) => {
+      const link = new namespace.elements.Link();
+      link.relation = 'terms-of-service';
+      link.href = termsOfService;
 
-    return link;
-  };
+      return link;
+    });
 
   const parseMember = R.cond([
     [hasKey('title'), parseString(context, name, true)],
-    [hasKey('termsOfService'), parseString(context, name, true)],
+    [hasKey('termsOfService'), parseTermsOfService],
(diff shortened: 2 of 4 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parseInfoObject.js
```diff
@@ -35,8 +35,17 @@ const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
...
+    const link = new namespace.elements.Link();
+    link.relation = 'terms-of-service';
+    link.href = terms;
+
+    return link;
+  };
+
   const parseMember = R.cond([
     [hasKey('title'), parseString(context, name, true)],
+    [hasKey('termsOfService'), parseString(context, name, true)],
```
marcofriso: I am not sure about the 3rd parameter, whether should be `true` or `false`

kylef: It should be `false`, it represents that `termsOfService` is optional and not required.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseInfoObject-test.js
```diff
@@ -178,4 +165,19 @@ describe('#parseInfoObject', () => {
...
+      termsOfService: 'http://example.com/terms/',
+    });
+
+    const parseResult = parse(context, info);
+    expect(parseResult.length).to.equal(1);
+
+    const termsOfService = parseResult.api.links.get(0);
+    expect(termsOfService.relation.toValue()).to.equal('terms-of-service');
+    expect(termsOfService.href.toValue()).to.equal('http://example.com/terms/');
+  });
```
kylef: Could we get a test to check what happens when the link is not a string?

Something along the lines of:

```patch
+    it('provides a warning when termsOfService is not a string', () => {
+      const info = new namespace.elements.Object({
+        title: 'My API',
+        version: '1.0.0',
+        termsOfService: 6,
+      });
+
+      const parseResult = parse(context, info);
+      expect(parseResult).to.contain.warning("'Info Object' 'termsOfService' is not a string");
+    });
```

I'd wrap these two test cases in `describe('#termsOfService', () => {`

packages/fury-adapter-oas3-parser/lib/parser/oas/parseInfoObject.js
```diff
@@ -58,6 +67,10 @@ function parseInfo(context, info) {
       api.title = info.get('title');
       api.attributes.set('version', info.get('version'));
 
+      if (info.get('termsOfService')) {
+        api.links.push(parseTermsOfService(info.get('termsOfService')));
```
kylef: The pattern in the code-base is generally in this block of code the everything is already parsed. I'd move the `parseTermsOfService` up a bit for consistency into something like:

```js
  const parseTermsOfService = pipeParseResult(namespace,
    parseString(context, name, false),
    getValue,
    (termsOfService) => {
      const link = new namespace.elements.Link();
      link.relation = 'terms-of-service';
      link.href = termsOfService;
      return link;
    });
```

Then parse the member as: `[hasKey('termsOfService'), parseTermsOfService],` then here it's just about connecting it up:

```js
      if (info.get('termsOfService')) {
        api.links.push(info.get('termsOfService'));
      }
```

packages/fury-adapter-oas3-parser/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Fury OAS3 Parser Changelog
 
+## Master
+
+### Enhancements
+
+- Adds support for the `info.termsOfService` schema extension.
```
kylef: I'd remove the wording regarding schema and extensions this isn't part of schema object or an extension. Perhaps something along the lines of:

```suggestion
- Adds support for `termsOfService` in 'Info Object'.
```

There's also a file called `STATUS.md` in this directory which tracks what features are supported. Could you take a look at that file and update `termsOfService`.

kylef: The STATUS file is at https://github.com/apiaryio/api-elements.js/blob/ed99d4967eb71383e9328cb3c56dcbd74e43262a/packages/fury-adapter-oas3-parser/STATUS.md#info-object

packages/fury-adapter-oas3-parser/lib/parser/oas/parseInfoObject.js
marcofriso: ok, change on next commit

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseInfoObject-test.js
marcofriso: ok, test added in next commit

packages/fury-adapter-oas3-parser/lib/parser/oas/parseInfoObject.js
marcofriso: refactor on next commit

packages/fury-adapter-oas3-parser/CHANGELOG.md
marcofriso: ok, change on next commit

PR #412, opened 2020-02-17
Enable source map tests for OAS 2 regression fixtures

kylef:
This change adds `.sourcemap.json` fixtures for every OAS 2 regression test, and the logic needed to test them.


kylef:
The test fixtures do not need to be fully analysed, these are the current behaviour, I am not changing behaviour in this PR only enabling these files.

marcofriso:
is it normal that some of the .json files in the test already contain source-maps?

kylef:
> is it normal that some of the .json files in the test folder already contain source-maps?


For annotations (warnings and errors) there will be source maps regardless.

PR #416, opened 2020-03-02
feat(oas2): add `contact` support

marcofriso:
Closes APIARY-6348

### DIFF
--- a/packages/fury-adapter-swagger/lib/parser.js
+++ b/packages/fury-adapter-swagger/lib/parser.js
@@ -372,35 +372,39 @@ class Parser {
             const { name, url, email } = this.swagger.info.contact;
 
             if (url) {
-              const link = new Link();
-              link.relation = 'contact';
-              link.href = url;
+              this.withPath('url', () => {
+                const link = new Link();
+                link.relation = 'contact';
+                link.href = url;
 
-              if (name) {
-                link.title = name;
-              }
+                if (name) {
+                  link.title = name;
+                }
 
-              if (this.generateSourceMap) {
-                this.createSourceMap(link, this.path);
-              }
+                if (this.generateSourceMap) {
+                  this.createSourceMap(link.href, this.path);
+                }
 
-              this.api.links.push(link);
+                this.api.links.push(link);
+              });
             }
 
             if (email) {
-              const link = new Link();
-              link.relation = 'contact';
-              link.href = email;
+              this.withPath('email', () => {
+                const link = new Link();
+                link.relation = 'contact';
+                link.href = `mailto:${email}`;
 
-              if (!url && name) {
-                link.title = name;
-              }
+                if (!url && name) {
+                  link.title = name;
+                }
 
-              if (this.generateSourceMap) {
-                this.createSourceMap(link, this.path);
-              }
+                if (this.generateSourceMap) {
+                  this.createSourceMap(link.href, this.path);
+                }
 
-              this.api.links.push(link);
+                this.api.links.push(link);
+              });
             }
           });
         }
--- a/packages/fury-adapter-swagger/lib/parser.js
+++ b/packages/fury-adapter-swagger/lib/parser.js
@@ -377,7 +377,7 @@ class Parser {
               link.href = url;
 
               if (name) {
-                link.title = name.clone();
+                link.title = name;
               }
 
               if (this.generateSourceMap) {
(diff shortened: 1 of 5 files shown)


packages/fury-adapter-swagger/lib/parser.js
```diff
@@ -366,6 +366,45 @@ class Parser {
...
+            const { Link } = this.namespace.elements;
+            const { name, url, email } = this.swagger.info.contact;
+
+            if (url) {
+              const link = new Link();
+              link.relation = 'contact';
+              link.href = url;
+
+              if (name) {
+                link.title = name.clone();
```
kylef:
```suggestion
                link.title = name;
```

Name here isn't an element so no need for `clone()`. It should be a `String` instance, the setter on `link.title` should wrap it in a `StringElement`

packages/fury-adapter-swagger/lib/parser.js
```diff
@@ -366,6 +366,45 @@ class Parser {
           });
         }
 
+        if (this.swagger.info.contact) {
+          this.withPath('contact', () => {
+            const { Link } = this.namespace.elements;
+            const { name, url, email } = this.swagger.info.contact;
+
+            if (url) {
```
kylef: I spotted below that the source maps for the URL and the email are the same, we can fix this by wrapping the URL and the email blocks in `this.withPath('url', () => {` and `this.withPath('email' => {` respectively.

```suggestion
            if (url) {
              this..withPath('url', () => {
                  ...
              });
            }
```

Afterwards, the source maps in the test fixture will show for the url and the email differently, as right now `this.path` is just set to `info.contact` so the source map is for the contact object not the specific value.

packages/fury-adapter-swagger/test/fixtures/info-contact.sourcemap.json
```diff
@@ -0,0 +1,162 @@
...
+            },
+            {
+              "element": "link",
+              "attributes": {
+                "relation": {
+                  "element": "string",
+                  "content": "contact"
+                },
+                "href": {
+                  "element": "string",
```
kylef:
```suggestion
                  "element": "string",
                  "attributes": { "sourceMap": ... }
```

packages/fury-adapter-swagger/lib/parser.js
```diff
@@ -366,6 +366,45 @@ class Parser {
...
+              const link = new Link();
+              link.relation = 'contact';
+              link.href = url;
+
+              if (name) {
+                link.title = name;
+              }
+
+              if (this.generateSourceMap) {
+                this.createSourceMap(link, this.path);
```
kylef:
```js
if (this.generateSourceMap) {
  this.withPath('url', () => {
    this.createSourceMap(link.href, this.path);
  });
}
```

We should apply the source map to the StringElement representing the HREF with the source map to the URL instead of the source map on the Link element

packages/fury-adapter-swagger/test/fixtures/info-contact.json
```diff
@@ -0,0 +1,66 @@
...
+            {
+              "element": "link",
+              "attributes": {
+                "relation": {
+                  "element": "string",
+                  "content": "contact"
+                },
+                "href": {
+                  "element": "string",
+                  "content": "support@example.com"
```
kylef: We should prefix this with `mailto:` to indicate it is an email link.

```suggestion
                  "content": "mailto:support@example.com"
```

kylef: https://en.wikipedia.org/wiki/Mailto

packages/fury-adapter-swagger/lib/parser.js
marcofriso: change on next commit

packages/fury-adapter-swagger/test/fixtures/info-contact.json
marcofriso: change on next commit

packages/fury-adapter-swagger/lib/parser.js
marcofriso: change on next commit, also for email

packages/fury-adapter-swagger/test/fixtures/info-contact.sourcemap.json
marcofriso: fix on next commit

PR #428, opened 2020-03-25
feat(core): add support for OAS3 Server object

marcofriso:
Closes APIARY-5855, APIARY-5856


packages/api-elements/lib/elements/Transition.js
```diff
@@ -116,6 +129,16 @@ class Transition extends ArrayElement {
...
+
+  /**
+   * @name host
+   * @type StringElement
+   * @memberof ParseResult.prototype
+   */
+  get host() {
+    return this.children
+      .filter(item => item.classes.contains('host'));
+  }
```
kylef:
```suggestion
```

I don't think we need this accessor, according to the design there won't be any hosts inside the content, only inside the `attributes` (handled above).

packages/api-elements/lib/elements/Resource.js
```diff
@@ -59,6 +72,16 @@ class Resource extends ArrayElement {
...
+
+  /**
+   * @name host
+   * @type StringElement
+   * @memberof ParseResult.prototype
+   */
+  get host() {
+    return this.children
+      .filter(item => item.classes.contains('host'));
+  }
```
kylef:
```suggestion
```

I don't think we need this accessor, according to the design there won't be any hosts inside the content, only inside the `attributes` (handled above).

packages/api-elements/lib/elements/Transition.js
marcofriso: ok, changed

packages/api-elements/lib/elements/Resource.js
marcofriso: ok, changed

marcofriso:
@kyle I am trying to add the support for the server object to the Fury adapter, but I am missing something, can you please have a check?
Thanks!

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
```diff
@@ -140,6 +143,14 @@ function parseOASObject(context, object) {
         }
       }
 
+      if (hosts) {
+        if (!hosts.isEmpty) {
+          api.push(new namespace.elements.Category(
+            hosts.content, { classes: ['hosts'] }
+          ));
+        }
```
kylef, line 151: I'd suggest moving this logic into `parseServersArray` so that the `parseServersArray` will return a Category element if there are `servers`. That way here you can reduce the logic to only incorporate merging the hosts category into the `api` and we have something such as:

```js
if (hosts) {
   api.push(hosts);
}
```

This can be appended to the `pipeParseResult` part of `parseServers` (array).

This is for a couple of reasons:

- To make `parseOpenAPIObject` simpler so it can deal with as little logic as possible.
- Make it easier to test, because you can test this behaviour in the `parseServersArray` without testing the rest of the OpenAPI Object behaviours.
- To encourage re-usability, when another component needs to parse servers using `parseServersArray`, it already returns the structure it will need instead of additional processing.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseServerObject.js
```diff
@@ -0,0 +1,49 @@
...
+  // [hasKey('variables'), R.compose(parseServerVariablesObject(context), getValue)], // NOT SUPPORTED YET
+  [hasKey('variables'), createUnsupportedMemberWarning(context.namespace, name)],
+  [isExtension, () => new context.namespace.elements.ParseResult()],
+  [R.T, createInvalidMemberWarning(context.namespace, name)],
+]);
+
+/**
+ * Parse the OpenAPI 'Server Object' (`#/server`)
+ * @see http://spec.openapis.org/oas/v3.0.3#server-object
+ * @returns ParseResult<Link>
```
kylef, line 28:
```suggestion
 * @returns ParseResult<Resource>
```

We're returning a resource element with this code.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseOpenAPIObject-test.js
```diff
@@ -47,6 +47,41 @@ describe('#parseOpenAPIObject', () => {
...
+      ],
+    });
+
+    const parseResult = parse(context, object);
+
+    expect(parseResult.length).to.equal(1);
+    expect(parseResult.api.title.toValue()).to.equal('My API');
+    expect(parseResult.api.length).to.equal(1);
+
+    const serversArray = parseResult.api.get(0);
```
kylef, line 75:
```suggestion
    const hostsCategory = parseResult.api.get(0);
```

As we're dealing in API Elements in these checks, I would use the API Elements terminology instead of "servers" the OAS terminology in this and below checks.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseOpenAPIObject-test.js
```diff
@@ -47,6 +47,41 @@ describe('#parseOpenAPIObject', () => {
...
+    expect(serversArray).to.be.instanceof(namespace.elements.Category);
+    expect(serversArray.classes.toValue()).to.deep.equal(['hosts']);
+    expect(serversArray.length).to.equal(2);
+
+    const firstServer = serversArray.get(0);
+    expect(firstServer).to.be.instanceof(namespace.elements.Resource);
+    expect(firstServer.attributes.content[0].toValue().value).to.equal('https://user.server.com/1.0');
```
kylef, line 82: I think you should be able to use the [`href`](https://github.com/apiaryio/api-elements.js/blob/ab47bebd0f421012f591982e1fe2512c41f6a9ab/packages/api-elements/lib/elements/Resource.js#L24) property on resource to simplify this code.

```suggestion
    expect(firstServer.href.toValue()).to.equal('https://user.server.com/1.0');
```

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServerObject-test.js
```diff
@@ -0,0 +1,79 @@
+const { Fury } = require('fury');
+const { expect } = require('../../chai');
+
+const parse = require('../../../../lib/parser/oas/parseServerObject');
+const Context = require('../../../../lib/context');
+
+const { minim: namespace } = new Fury();
+
+describe('#parseHostsObject', () => {
```
kylef, line 9:
```suggestion
describe('#parseServerObject', () => {
```

I think this has been renamed.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServerObject-test.js
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    it('warns when server object does not contain URL', () => {
+      const server = new namespace.elements.Object({
+      });
+
+      const parseResult = parse(context)(server);
+      expect(parseResult.length).to.equal(1);
+      expect(parseResult).to.contain.warning("'Server Object' is missing required property 'url'");
+    });
```
kylef, line 49: I think we're missing a test case for testing what happens when `url` is not a string. This is similiar to one of the below tests such as `warns when description is not a string`.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServersArray-test.js
```diff
@@ -0,0 +1,74 @@
...
+        url: 'https://user.server.com/1.0',
+        description: 'The production API server',
+      },
+    ]);
+
+    const parseResult = parse(context, server);
+
+    expect(parseResult.length).to.equal(1);
+
+    const servers = parseResult.get(0);
```
kylef, line 36:
```suggestion
    const hosts = parseResult.get(0);
```

Perhaps we can call this `hosts` to match API Elements terminology as the `parse` method returns API Elements.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServersArray-test.js
```diff
@@ -0,0 +1,74 @@
...
+
+    expect(servers).to.be.instanceof(namespace.elements.Array);
+    expect(servers.length).to.equal(1);
+    expect(servers.get(0)).to.be.instanceof(namespace.elements.Resource);
+    expect(servers.get(0).meta.content[0].content.value.content[0].content).to.equal('host');
+    expect(servers.get(0).meta.content[1].content.value.content).to.equal('The production API server');
+    expect(servers.get(0).attributes.content[0].content.value.content).to.equal('https://user.server.com/1.0');
```
kylef, line 43: This check is very specific to the structure, I think it can be rewritten (including the following two lines) to use common accessors such as `classes`, `href`, `description`. I would try and avoid using `content` directly.

I'm not sure if this will work, but its atleast a clearer direction to try and refactor to:

```suggestion
    const host = servers.get(0);
    expect(host.classes.toValue()).to.deep.equal(['host']);
    expect(host.description.toValue()).to.equal('The production API server');
    expect(host.href.toValue()).to.equal('https://user.server.com/1.0');
```

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServersArray-test.js
```diff
@@ -0,0 +1,74 @@
...
+    expect(servers).to.be.instanceof(namespace.elements.Array);
+    expect(servers.length).to.equal(2);
+    expect(servers.get(0)).to.be.instanceof(namespace.elements.Resource);
+    expect(servers.get(0).meta.content[0].content.value.content[0].content).to.equal('host');
+    expect(servers.get(0).attributes.content[0].content.value.content).to.equal('https://user.server.com/1.0');
+
+    expect(servers.get(1)).to.be.instanceof(namespace.elements.Resource);
+    expect(servers.get(1).meta.content[0].content.value.content[0].content).to.equal('host');
+    expect(servers.get(1).meta.content[1].content.value.content).to.equal('The production API server');
+    expect(servers.get(1).attributes.content[0].content.value.content).to.equal('https://user.server.com/2.0');
```
kylef, line 72: You should be able to incorporate similair feedback to my previous comment and make these checks simpler using `classes`, `href`, `description` properties.

packages/fury-adapter-oas3-parser/lib/parser/oas/parseHostsObject.js
```diff
@@ -0,0 +1,57 @@
...
+        resource.hrefVariables = variables;
+      }
+
+      resource.href = hostObject.getValue('url');
+
+      hosts.push(resource);
+    });
+  });
+
+module.exports = parseHostsObject;
```
kylef, line 57: I think this file can be removed as this is now `parseServerObject.js`

packages/fury-adapter-oas3-parser/lib/parser/oas/parseServerObject.js
```diff
@@ -0,0 +1,49 @@
...
+  R.unless(isObject, createWarning(context.namespace, `'${name}' is not an object`)),
+  parseObject(context, name, parseMember(context), requiredKeys, [], true),
+  (object) => {
+    const resource = new context.namespace.elements.Resource();
+
+    resource.classes.push('host');
+
+    if (object.hasKey('description')) {
+      const description = object.getValue('description');
+      resource.description = description;
```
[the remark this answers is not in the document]
kylef: I forgot to update this and the below comment with the explanation so here goes.

`getValue` property on an `ObjectElement`/`ArrayElement` will return a primitive value, in this case a JavaScript String.

`get` will return the value as an element, which in this case is a `StringElement`.

Effectively, `getValue` is implemented like following:

```js
function getValue(indexOrKey) {
  const item = this.get(indexOrKey);

  if (item) {
    return item.toValue();
  }

  return undefined;
}
```

We want to set resource's description and href using the `StringElement`, because it can contain source map information which is lossed when an element is converted to a string.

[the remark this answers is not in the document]
marcofriso: I see the point, changed

packages/fury-adapter-oas3-parser/lib/parser/oas/parseServerObject.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    resource.href = object.getValue('url');
```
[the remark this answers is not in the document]
marcofriso: changed

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
marcofriso: ok, changed

packages/fury-adapter-oas3-parser/lib/parser/oas/parseServerObject.js
marcofriso: right, changed

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
marcofriso: @kylef 
I have a doubt about this because if in `parseServerArray` I change the function

```
function parseServersArray(context, element) {

  const parseServers = pipeParseResult(namespace,
    parseArray(context, name, R.curry(parseServerObject)(context)));
  return parseServers(element);
}
```
to include this part, it will become something like
```
function parseServersArray(context, element) {

  const parseServers = pipeParseResult(namespace,
    parseArray(context, name, R.curry(parseServerObject)(context)));

  return new namespace.elements.Category(
    parseServers(element), { classes: ['hosts'] }
  );
}
```
and, despite it works, the Categories are one level down, with `content` level that adds in between and a `content: parseResult` within the ServerArray. Could you please tell me how to change this function to not mess up the structure?

packages/fury-adapter-oas3-parser/lib/parser/oas/parseServerObject.js
marcofriso: right, change on next commit

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseOpenAPIObject-test.js
marcofriso: ok changed;
it is easy to confuse the terminology while writing parsing methods.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServerObject-test.js
marcofriso: all right I will do that

packages/fury-adapter-oas3-parser/lib/parser/oas/parseHostsObject.js
marcofriso: removed

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServersArray-test.js
marcofriso: word replaced for all the cases that are the result of the parser

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServersArray-test.js
marcofriso: replaced

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServersArray-test.js
marcofriso: all replaced and working, I have to study the elements to shorten the structure during tests and functions.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServerObject-test.js
marcofriso: returned to old denomination

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
kylef: @marcofriso you should be able to add another function argument to `pipeParseResult`, the output of `parseArray` on success will be passed to the next function before returning from `pipeParseResult`. This is similiar to what you've done in `parseServerObject`. I think something along the lines of:

```js
function parseServersArray(context, element) {

  const parseServers = pipeParseResult(namespace,
    parseArray(context, name, R.curry(parseServerObject)(context)),
    (array) => {
      const category = new Category(array.content);
      category.classes.push('hosts');
      return category;
    });
  return parseServers(element);
}
```

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseOpenAPIObject-test.js
marcofriso: changed

packages/fury-adapter-oas3-parser/lib/parser/oas/parseOpenAPIObject.js
kylef: `pipeParseResult` behaves similiar to [Rambda pipe](https://ramdajs.com/docs/#pipe), in which it performs left-to-right function composition. The result of a left function is passed to the right function. In `pipeParseResult` that only happens if the parse result that `parseArray` returns contains a value so that it can "error-out".

marcofriso: good,
in the end the function turns like this:

```js
function parseServersArray(context, element) {

  const parseServers = pipeParseResult(namespace,
    parseArray(context, name, R.curry(parseServerObject)(context)),
    array => new namespace.elements.Category(
      array.content, { classes: ['hosts'] }
    ));

  return parseServers(element);
```

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServerObject-test.js
kylef: I think this is still valid, the name here should match the function we are testing, which is `parseServerObject`, from:

```js
const parse = require('../../../../lib/parser/oas/parseServerObject');
```

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServersArray-test.js
```diff
@@ -0,0 +1,77 @@
+const { Fury } = require('fury');
+const { expect } = require('../../chai');
+const parse = require('../../../../lib/parser/oas/parseServersArray');
+const Context = require('../../../../lib/context');
+
+const { minim: namespace } = new Fury();
+
+describe('#parseHostsObject', () => {
```
kylef, line 8:
```suggestion
describe('#parseServersArray', () => {
```

To match the function we are testing.

marcofriso: changed

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServerObject-test.js
marcofriso: yes, changed, sorry

marcofriso:
@kylef As https://github.com/apiaryio/api-elements/pull/62 has been merged I am merging also this PR

PR #432, opened 2020-04-06
Marco/oas3 server variable

marcofriso:
Closes APIARY-5856, APIARY-6481


packages/fury-adapter-oas3-parser/lib/parser/oas/parseServerObject.js
```diff
@@ -41,6 +42,10 @@ const parseServerObject = context => pipeParseResult(context.namespace,
 
     resource.href = object.get('url');
 
+    if (object.hasKey('variables')) {
+      resource.push(object.get('variables'));
```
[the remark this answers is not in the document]
marcofriso: Addressed in the last commit

packages/fury-adapter-oas3-parser/lib/parser/oas/parseServerVariablesArray.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+const parseServerVariablesArray = (context, element) => {
+  const { namespace } = context;
+
+  const parseServerVariables = pipeParseResult(namespace,
+    parseArray(context, name, R.curry(parseServerVariableObject)(context)),
+    array => new namespace.elements.HrefVariables(array.content));
+
+  return parseServerVariables(element);
+};
```
kylef, line 25: I think perhaps you've mis-understood the OpenAPI spec as I don't think there is any server variables array defined. The `variables` property of "Server Object" contains an object (key-value) pair. For example:

```yaml

servers:
  - url: https://{username}.gigantic-server.com:{port}/{basePath}
    description: The production API server
    variables:
      username:
        # note! no enum here means it is an open value
        default: demo
        description: this value is assigned by the service provider, in this example `gigantic-server.com`
      port:
        enum:
          - '8443'
          - '443'
        default: '8443'
      basePath:
        # open meaning there is the opportunity to use special base paths as assigned by the provider, default is `v2`
        default: v2
```

"variables" contains an object with three keys "username", "port", and "basePath".

kylef: `parseServerVariablesArray` should perhaps become `parseServerVariablesObject` which accepts an object such as:

```yaml
username:
  default: x
port:
  default '80'
```

and returns `HrefVariables`, such as:

```json
{
  "element": "hrefVariables",
  "content": [
    {
      "element": "member",
      "content": {
        "key": {
          "element": "string",
          "content": "username"
        },
        "value": { ... }
      }
    },
    {
      "element": "member",
      "content": {
        "key": {
          "element": "string",
          "content": "port"
        },
        "value": { ... }
      }
    }
  ]
}
```

marcofriso: I see;
`serverVariableArray` was just a way to bring the server variables to the server object

packages/fury-adapter-oas3-parser/lib/parser/oas/parseServerVariableObject.js
```diff
@@ -0,0 +1,50 @@
...
+const parseObject = require('../parseObject');
+const parseString = require('../parseString');
+const parseEnum = require('../parseEnum');
+const pipeParseResult = require('../../pipeParseResult');
+
+const name = 'Server Variable Object';
+const requiredKeys = ['default'];
+
+const parseMember = context => R.cond([
+  [hasKey('enum'), R.compose(parseEnum(context, name), getValue)],
```
kylef, line 18: The OAS spec specifies that `enum` for a Server Variable Object should be an array of string which this implementation is not enforcing. This is not the same rule for Schema Object. I think we will need to make another `parseEnum` specific for Server Variable Object to warn and discard (or coerce) values which are not strings in the enum. It could make use of the shared `parseEnum` but add additional checks for the contents of the array.

That said, it probably isn't worth the effort I think it will be simpler to just handle the enums here because `parseArray` accepts a sub-parser (for example the string parser):

```js
const parseEnum = pipeParseResult(
  parseArray(context, `${name}' 'enum`, parseString(...)),
  (array) => {
    const element = new context.namespace.elements.Enum(array);
    element.enumerations.forEach(value => value.attributes.set('typeAttributes', ['fixed']));
    return element;
  }
);
```

We should have a test case for parseServerVariableObject's when the enum contains non strings. For example if the source was:

```js
['a', 1, 'b', true]
```

The resulting enum should be of `'a'` and `'b'`, and there should be two warnings, one for the value `1` and `true`.

marcofriso:
@kylef I realise that the Server Array has been merged as part of the other branch as OAS3 `servers` looks like this:
```
{
  "servers": [
    {
      "url": "https://development.gigantic-server.com/v1",
      "description": "Development server"
    },
    {
      "url": "https://staging.gigantic-server.com/v1",
      "description": "Staging server"
    },
  ]
}
```
all comments of this branch until now have been addressed

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServerObject-test.js
```diff
@@ -88,4 +88,56 @@ describe('#parseServerObject', () => {
...
+
+      const parseResult = parse(context)(server);
+      expect(parseResult).to.not.contain.annotations;
+      expect(parseResult.length).to.equal(1);
+
+      const resource = parseResult.get(0);
+      expect(resource).to.be.instanceof(namespace.elements.Resource);
+
+      const { hrefVariables } = resource;
+      const firstHrefVariable = hrefVariables.content.content[0];
```
opichals, line 127: @kylef @marcofriso  This seems inconsistent with what we have in other `Resource` elements where the hrefVariables directly contain an array of `Member` elements.

marcofriso: https://github.com/apiaryio/api-elements.js/pull/472

PR #435, opened 2020-04-16
Add support for OAS 3 default responses

kylef:
Default responses is an OpenAPI feature which allows you to specify a response that doesn't have a fixed status code. For example, if you want to specify a particular structure will be the response, but the status code can be anything.

This translates to no status code in the API Elements parse result, there is no example status code. Tooling such as Dredd (Gavel), ApiaryUI, Styleguides already handle cases where status code is missing. As for v3 docs and Apiary mock server, they default to 200 in this case, which is an acceptable solution. Ideally v3 docs could show no status code, but as OAS 3 support is experimental this is an acceptable comprimise to me.

### DIFF
--- a/packages/fury-adapter-oas3-parser/CHANGELOG.md
+++ b/packages/fury-adapter-oas3-parser/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Fury OAS3 Parser Changelog
 
+## Master
+
+### Enhancements
+
+- Support for "default" response status codes.
+
 ## 0.10.2 (2020-03-16)
 
 ### Enhancements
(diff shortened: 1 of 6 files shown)


kylef:
> default status code could be any 3 digits number, even if the number doesn't make sense?

@marcofriso I'm not sure I fully understand the question. Status code's can be either a number (for example `200`), `default`, or a range `2XX`, `4XX`.

With this PR we support fixed number like `200` and `default` but not ranges.

marcofriso:
@kylef I meant if it would work if the number was `001` or `999` for example

kylef:
Yes it will, there's existing test cases for that.

PR #441, opened 2020-04-24
Marco/support servers in path item

marcofriso:
Closes APIARY-6503

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parsePathItemObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parsePathItemObject.js
@@ -132,7 +132,6 @@ function parsePathItemObject(context, member) {
     [isHttpMethodKey, parseOperationObject(context, member.key)],
 
     // FIXME Parse $ref
-    // FIXME Parse servers
 
     [isUnsupportedKey, createUnsupportedMemberWarning(namespace, name)],
 
@@ -152,10 +151,7 @@ function parsePathItemObject(context, member) {
       resource.href = hrefFromParameters(member.key, parameters);
       resource.hrefVariables = hrefVariablesFromParameters(namespace, parameters);
 
-      const hosts = pathItem.get('servers');
-      if (hosts) {
-        resource.push(hosts);
-      }
+      resource.hosts = pathItem.get('servers');
 
       const summary = pathItem.get('summary');
       if (summary) {
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parsePathItemObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parsePathItemObject.js
@@ -9,12 +9,13 @@ const parseObject = require('../parseObject');
 const parseString = require('../parseString');
 const parseCopy = require('../parseCopy');
 const parseParameterObjects = require('./parseParameterObjects');
+const parseServersArray = require('./parseServersArray');
 const parseOperationObject = require('./parseOperationObject');
 const pipeParseResult = require('../../pipeParseResult');
 
 const name = 'Path Item Object';
 const httpMethods = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace'];
-const unsupportedKeys = ['$ref', 'servers'];
+const unsupportedKeys = ['$ref'];
 
 const isHttpMethodKey = R.anyPass(R.map(hasKey, httpMethods));
 const isUnsupportedKey = R.anyPass(R.map(hasKey, unsupportedKeys));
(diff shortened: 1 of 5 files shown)


packages/fury-adapter-oas3-parser/lib/parser/oas/parsePathItemObject.js
```diff
@@ -127,6 +128,7 @@ function parsePathItemObject(context, member) {
     [hasKey('summary'), parseString(context, name, false)],
     [hasKey('description'), parseCopy(context, name, false)],
     [hasKey('parameters'), R.curry(parseParameters)(context, member.key)],
+    [hasKey('servers'), R.compose(parseServersArray(context, name), getValue)],
```
kylef, line 131: There's a comment `// FIXME Parse servers` a few lines below, could you remove that please.

packages/fury-adapter-oas3-parser/lib/parser/oas/parsePathItemObject.js
```diff
@@ -150,6 +152,11 @@ function parsePathItemObject(context, member) {
       resource.href = hrefFromParameters(member.key, parameters);
       resource.hrefVariables = hrefVariablesFromParameters(namespace, parameters);
 
+      const hosts = pathItem.get('servers');
+      if (hosts) {
+        resource.push(hosts);
+      }
+
```
kylef, line 159: The host resources should be stored in the "hosts" attribute as per the design in https://github.com/apiaryio/api-elements/commit/2f9cebfc7d3b3a710172aeb291c4533706352377.

You've added a setter for [`hosts` on the Resource element](https://github.com/apiaryio/api-elements.js/blob/10841673ef9cd813ac9c1db7a57c1bc71765f5a4/packages/api-elements/lib/elements/Resource.js#L28) which I think the intention is that it can be used with something along the lines of:

```js
resource.hosts = pathItem.get('servers');
```

Since `get` should return undefined, I'd imagine that the setter for hosts is a no-op so you can omit checks.

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServersArray-test.js
```diff
@@ -15,21 +15,23 @@ describe('#parseServersArray', () => {
   it('warns when it is not an array', () => {
     const servers = new namespace.elements.Object();
 
-    const parseResult = parse(context, servers);
+    const name = 'parent-name';
```
kylef, line 18: You could perhaps define this at the top of the file as it's repeated a few times.

packages/fury-adapter-oas3-parser/lib/parser/oas/parsePathItemObject.js
marcofriso: removed

packages/fury-adapter-oas3-parser/test/unit/parser/oas/parseServersArray-test.js
marcofriso: moved to top

packages/fury-adapter-oas3-parser/lib/parser/oas/parsePathItemObject.js
marcofriso: @kylef 
> You've added a setter for hosts on the Resource element which I think the intention is that it can be used with something along the lines of:

is the `hosts` already added in the server array classes?
https://github.com/apiaryio/api-elements.js/blob/10841673ef9cd813ac9c1db7a57c1bc71765f5a4/packages/fury-adapter-oas3-parser/lib/parser/oas/parseServersArray.js#L23

> Since get should return undefined, I'd imagine that the setter for hosts is a no-op so you can omit checks.

Do you mean that I can remove some of the test?

kylef: > Do you mean that I can remove some of the test?

I meant the `if (hosts)` check is perhaps not needed (providing setter works as intended with setting undefined`.

kylef: > isn't the hosts already added in the server array classes?

It is, but I guess it isn't needed, in that case you could set the `hosts` as array element via something like:

```js
const hosts = pathItem.get('servers');
if (hosts) {
  resource.hosts = hosts.content;
}
```

That way the `hosts` class and category element is discarded here.

marcofriso: ok so in `parsePathItemObject.js` I set the import of the ServersArray as such:

```
resource.hosts = pathItem.get('servers');
```

If you want me to remove the `hosts` `classes` from `parseServersArray.js` tell me

PR #444, opened 2020-04-24
feat(oas3): add support to 'servers' in 'operation' object

marcofriso:
Closes APIARY-6502

### DIFF
--- a/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
+++ b/packages/fury-adapter-oas3-parser/lib/parser/oas/parseOperationObject.js
@@ -13,6 +13,7 @@ const parseObject = require('../parseObject');
 const parseString = require('../parseString');
 const parseResponsesObject = require('./parseResponsesObject');
 const parseParameterObjects = require('./parseParameterObjects');
+const parseServersArray = require('./parseServersArray');
 const parseRequestBodyObject = require('./parseRequestBodyObject');
 const parseSecurityRequirementsArray = require('./parseSecurityRequirementsArray');
 const parseReference = require('../parseReference');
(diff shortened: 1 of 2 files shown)


marcofriso:
tests are succeeding locally

kylef:
@marcofriso looking at the output on CI it looks like the integration tests fail:

```
  3) components
       Path Item Object
         handles multiple references to same parameter:

      Parse Result does not match '/home/circleci/repo/packages/fury-adapter-oas3-parser/test/integration/fixtures/components/path-item-object-parameters-multiple.json' fixture
      + expected - actual

                         "element": "string",
                         "content": "List all pets"
                       }
                     },
      -              "attributes": {},
                     "content": [
                       {
                         "element": "httpTransaction",
                         "content": [
```

It looks like setting attributes as undefined causes them to be deserialized as an empty object (which I don't think minim should do that).

Perhaps the best route of action would be to make it so that Minim doesn't serialize empty attributes under this case (in particular the line at https://github.com/refractproject/minim/blob/5e94e18fafea47ac8edb1dcae617ec9bf8dfa753/lib/serialisers/JSONSerialiser.js#L63 could be updated to also check that the object isn't empty).

Do you want to take a look at updating minim to do that? I'd make `meta` behave the same way while we're at it.

I think a test similiar to https://github.com/refractproject/minim/blob/5e94e18fafea47ac8edb1dcae617ec9bf8dfa753/test/serialisers/JSONSerialiser-test.js#L163 but more along the lines of "serialises an element without empty attributes" where test looks something like:

```
element.attributes.set('thread', undefined);
```

kylef:
Actually, I misread the code. There is already a length check, but since the length is non-zero it continues, just that one of the values is empty. https://github.com/refractproject/minim/blob/5e94e18fafea47ac8edb1dcae617ec9bf8dfa753/lib/serialisers/JSONSerialiser.js#L30 as this is serialise, not deserialise. So perhaps the mix is more along the lines of making serialiseObject (https://github.com/refractproject/minim/blob/5e94e18fafea47ac8edb1dcae617ec9bf8dfa753/lib/serialisers/JSONSerialiser.js#L138) return undefined under the case there isn't any values in the result.

marcofriso:
@kylef changes on Minim
→ https://github.com/refractproject/minim/pull/258

kylef:
@marcofriso https://github.com/apiaryio/api-elements.js/pull/448 updates to a new minim which solves that problem, we can merge that and then rebase this PR on master to bring it in.

PR #445, opened 2020-04-24
docs(oas3): indicate updated server support

marcofriso:


kylef:
@marcofriso is this PR correct? It looks like the same as https://github.com/apiaryio/api-elements.js/pull/444 which adds servers to operation object.

marcofriso:
> @marcofriso is this PR correct? It looks like the same as #444 which adds servers to operation object.

@kylef no, something went wrong, and also my backup disk doesn't work, but it was nothing too complicate, I am going to fix it or write it again

packages/fury-adapter-oas3-parser/STATUS.md
```diff
@@ -24,6 +24,38 @@ Key:
...
+| license | ✓ |
+| version | ✓ |
+
+## Contact Object
+
+| name | ✓ |
+| url | ✓ |
+| email | ✓ |
+
+## Licence Object
```
kylef, line 42:
```suggestion
## License Object
```

License is spent differently in the spec: https://spec.openapis.org/oas/v3.0.2#license-object

packages/fury-adapter-oas3-parser/STATUS.md
```diff
@@ -24,6 +24,38 @@ Key:
...
+
+## Server Object
+
+| url | ✓ |
+| description | ✓ |
+| variables | ✓ |
+
+## Server Variable Object
+
+| enum | ✓ |
```
kylef, line 55:
```suggestion
| Field Name | Support |
|:--|:--|
| enum | ✓ |
```

We're missing the table header rows on all the introduced objects here, could you update them to be consistent with the rest.

packages/fury-adapter-oas3-parser/STATUS.md
marcofriso: sure, changed

packages/fury-adapter-oas3-parser/STATUS.md
marcofriso: changed

Issue #32, opened 2018-12-07
Package Naming


kylef:
We should settle on the package naming going forward along with how we will package adapters up as we can also use NPM Scoped packages https://docs.npmjs.com/misc/scope (`@apielements/openapi` etc). We have the following packages at the moment:

- `minim-api-description`
- `minim-parse-result`
- `fury`
- `fury-cli`
- `fury-adapter-apib-parser`
- `fury-adapter-apiary-blueprint-parser`
- `fury-adapter-swagger`
    - `swagger-zoo`
- `fury-adapter-apib-serializer`

Perhaps the adapters could become something like `@apielements/openapi2-parser`, `@apielements/openapi3-parser`, `@apielements/apiblueprint-parser`. We could offer `@apielements/cli` and some package containing the base (fury.js) and namespace.

We could also offer a root level package `apielements` which provides everything, it could provide all of the adapters that we provide along side the CLI tool. That way consumers like Dredd (@honzajavorek), and Documentation (@char0n) have a single dependency use `apielements` which has all of the provided adapters already included without them having to deal with each adapter.

I know @honzajavorek had some gripes with using a dash in `api-elements` package so we can settle that too. I'd appreciate any feedback on the naming convention here.

/c @pksunkara

honzajavorek:
Yup, I like `apielements` more than `api-elements`, the dash seems to be redundant as even without it the name reads well, but types better.

I'd perhaps keep the CLI separated from the rest, i.e. the `apielements` package wouldn't contain it by default. I believe the use cases are different and the CLI is more of a convenience tool for developers than something one would need to have in their production dependencies. Otherwise it seems to be a good idea:

- `minim-api-description` → ?
- `minim-parse-result` → ?
- `fury` → `@apielements/core`
- `fury-cli` → `@apielements/cli`
- `fury-adapter-apib-parser` → `@apielements/apiblueprint-parser`
    - `mson-zoo` → `@apielements/mson-examples` ?
- `fury-adapter-apiary-blueprint-parser` → `@apielements/apiaryblueprint-parser`
- `fury-adapter-swagger` → `@apielements/openapi2-parser`
    - `swagger-zoo` → `@apielements/openapi2-examples`
- `fury-adapter-apib-serializer` → `@apielements/apiblueprint-serializer`

...and

`@apielements/*` - `@apielements/cli` = `apielements`

Almad:
[...]

PR #452, opened 2020-04-29
Rename apib-parser

kylef:
As per https://github.com/apiaryio/api-elements.js/issues/32 (in particular https://github.com/apiaryio/api-elements.js/issues/32#issuecomment-621226696), we are renaming Fury into API Elements. I will be changing each adapter independently in separate PRs (to try and keep them small and reviewable).


kylef:
As for Snyk, I will need to update the configuration on Snyk to make that pass, so will do once this is approved and ready for merge.

packages/apib-parser/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-# Fury API Blueprint Parser Changelog
+# API Elements: API Blueprint Parser Changelog
+
+## Master
```
kylef, line 3: ❓ I was planning to release this as 0.18.1 so that the "same" version of this adapter can be used with both package names. Not sure what to do in changelog though.

kylef: Perhaps its clearer if I make this as a `0.19.0` release with the new name.

kylef: @opichals ok, i've pushed 541dd8e4 which bumps the version of this adapter to 0.19.0.

PR #156, opened 2019-02-12
Added partial support for the `example` in Parameter Object

lulco:
Added support for `example` in parameters:

```yaml
      parameters:
        -
          name: code
          in: query
          required: true
          description: 'Code to check'
          example: ABCDEFG
          schema:
            type: string
```

Not working with
```yaml
      parameters:
        -
          name: code
          in: query
          required: true
          description: 'Code to check'
          schema:
            type: string
            example: ABCDEFG
```
yet.

This helps me to (partially) resolve issue https://github.com/apiaryio/dredd/issues/1236

Not sure, if this is correct, it is my first contribution. So please, feel free to correct my code.

PR #459, opened 2020-05-05
Add support for schema example in OAS3 parameters

kmdsbng:
According to OAS3 Specification, an example element can be included in a schema object.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schemaObject

I fixed to use example value under the schema object under the parameter object when it has set.

I'm using Stoplight Studio and I'm having the following problem, so I'd like you to import this pull request.

problem: When I set a example value of Query parameter in Stoplight Studio, it will output the value to parameters.schema.example in the OAS3 yaml file. If this file is read by dredd, an error will be displayed as "example" is not set. 

I referenced #156 to make this pull request.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseParameterObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseParameterObject.js
@@ -1,6 +1,6 @@
 const R = require('ramda');
 const {
-  isExtension, hasKey, hasValue, getValue,
+  isObject, isExtension, hasKey, hasValue, getValue,
 } = require('../../predicates');
 const {
   createError,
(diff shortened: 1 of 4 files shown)


kmdsbng:
coverall ci has failed. But I had no idea for fix this problem.
Can I do anything for fix this?

https://circleci.com/gh/apiaryio/api-elements.js/6742?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

```
> nyc report --reporter=text-lcov | coveralls


/home/circleci/repo/node_modules/coveralls/bin/coveralls.js:19
      throw err;
      ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
(Use `node --trace-uncaught ...` to show where the exception was thrown)
```

PR #460, opened 2020-05-06
Don't limit scripts to particular scope

kylef:
Some packages are in @apielements and this filter for `--scope '*'` seems to prevent them from being executed in test commands.

This can be quickly confirmed via comparing the test run times between master and this branch (and before I renamed some packages).

Screenshot 2020-05-06 at 11 20 21

Screenshot 2020-05-06 at 11 20 32

The output can also be analysed to see that some packages like openapi3-parser were not tested.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,9 @@
   "scripts": {
     "release": "lerna publish",
     "release:force": "lerna publish --force-publish",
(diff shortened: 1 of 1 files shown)


kylef:
The coverage report is a big indication of this too.

### Before

misses many files

Screenshot 2020-05-06 at 11 31 07


### After

Screenshot 2020-05-06 at 11 31 15

PR #461, opened 2020-05-12
Fix handling of schema object's without type

kylef:
In a schema object which does not specify a `type`, any type may be permitted as a value. API Elements doesn't have an element type that represens "any", semantically it is equivilent to an enum of any permitted types (3 primitive types of boolean, number, string and 2 collection types: object and array).

This is the same as the behaviour in the OpenAPI 3 parser (this is the OAS 2 equivilent of https://github.com/apiaryio/api-elements.js/pull/135).

For example, the following schema:

```yaml
items: { type: string }
properties:
  name: { type: string }
```

Which is semantically equivilent to the following MSON:

+ enum
  + (object)
    + name (string)
  + (array[string], fixed-type)
  + (boolean)
  + (string)
  + (number)

Would allow the following values:

```json
["a", "b"]
```

```json
"hello world"
```

```json
{
  "name": "doe"
}
```

However, the following values are not permitted:

```json
[1, true]
```

```json
{
  "name": true
}
```

There is an existing behaviour which has been altered, in the past we "coered" a schema object which contains a properties key to be of only type object. This is incorrect behaviour which goes against the above support.

### DIFF
--- a/packages/openapi2-parser/lib/schema.js
+++ b/packages/openapi2-parser/lib/schema.js
@@ -203,11 +203,6 @@ class DataStructureGenerator {
           return uniqueTypes[0];
         }
       }
-
-      if (schema.properties) {
-        // Assume user meant object
-        return 'object';
-      }
     }
 
     return schema.type;
@@ -271,6 +266,15 @@ class DataStructureGenerator {
       element = new typeGeneratorMap[type]();
     } else if (type) {
       throw new Error(`Unhandled schema type '${type}'`);
+    } else {
+      element = new this.minim.elements.Enum();
+      element.enumerations = [
+        new this.minim.elements.String(),
+        new this.minim.elements.Number(),
+        new this.minim.elements.Boolean(),
+        this.generateArray(schema),
+        this.generateObject(schema),
+      ];
     }
 
     if (element) {
--- a/packages/openapi2-parser/test/fixtures/circular-example.json
+++ b/packages/openapi2-parser/test/fixtures/circular-example.json
@@ -405,28 +405,25 @@
                         "content": "company"
                       },
                       "value": {
-                        "element": "object",
-                        "content": [
-                          {
-                            "element": "member",
-                            "attributes": {
-                              "typeAttributes": {
-                                "element": "array",
-                                "content": [
-                                  {
-                                    "element": "string",
-                                    "content": "optional"
-                                  }
-                                ]
-                              }
-                            },
-                            "content": {
-                              "key": {
-                                "element": "string",
-                                "content": "data"
+                        "element": "enum",
+                        "attributes": {
+                          "enumerations": {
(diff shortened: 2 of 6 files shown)


packages/openapi2-parser/test/fixtures/circular-example.json
```diff
@@ -405,28 +405,25 @@
                         "content": "company"
                       },
                       "value": {
-                        "element": "object",
```
kylef, line 408: This and the other JSON fixtures are results of the behaviour change for when a schema object which contains properties was coered as an object.

packages/openapi2-parser/test/schema-test.js
```diff
@@ -754,14 +786,15 @@ describe('JSON Schema to Data Structure', () => {
     it('produces empty array element with empty items', () => {
       const schema = {
         type: 'array',
-        items: {},
+        items: { type: 'string' },
```
kylef, line 789: As lack of type is now handled differently this test would be broken. I've explicitly specified a strng type so that the test assertions are clearer (otherwise the value is enum of 5 types).

packages/openapi2-parser/lib/schema.js
```diff
@@ -203,11 +203,6 @@ class DataStructureGenerator {
           return uniqueTypes[0];
         }
       }
-
-      if (schema.properties) {
```
freaz, line 207: this removes only assumption we made right?

packages/openapi2-parser/lib/schema.js
```diff
@@ -271,6 +266,15 @@ class DataStructureGenerator {
...
       throw new Error(`Unhandled schema type '${type}'`);
+    } else {
+      element = new this.minim.elements.Enum();
+      element.enumerations = [
+        new this.minim.elements.String(),
+        new this.minim.elements.Number(),
+        new this.minim.elements.Boolean(),
+        this.generateArray(schema),
+        this.generateObject(schema),
+      ];
```
freaz, line 277: And here we create enum with all datatypes. Also in case schema has for eg. items it will create array with items of given type.

kylef: Exactly, we don't support other validation keywords for the primitive types so we can create an empty primitive element (boolean, string, number) in those cases.

In the cases of array and object we support more keywords so there's handling of using those existing functions which process the items/properties/required and similiar keywords.

packages/openapi2-parser/lib/schema.js
kylef: That's right

kylef:
> Do you plan to increase major version?

Yes I think that makes sense as its possibly a breaking change in terms that missing type used to be coereced as an object (incorrectly). So 0.31.0 because we're pre 1.0.

PR #464, opened 2020-05-12
Required schema object keys which are not in properties

kylef:
This is the same bug that existed in the OAS 2 parser at https://github.com/apiaryio/api-elements.js/pull/462, see the related PR for description.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
@@ -39,13 +39,13 @@ function constructObjectStructure(namespace, schema) {
 
   const required = schema.get('required');
   if (required) {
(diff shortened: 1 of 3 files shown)


packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -39,13 +39,19 @@ function constructObjectStructure(namespace, schema) {
...
   const required = schema.get('required');
   if (required) {
-    required.forEach((key) => {
-      const member = element.getMember(key.toValue());
-
-      if (member) {
-        member.attributes.set('typeAttributes', ['required']);
+    const findOrCreateMember = R.either(
+      element.getMember.bind(element),
+      (key) => {
```
opichals, line 44: nit: Perhaps extract this into a named function?

PR #468, opened 2020-05-20
Add an OpenAPI 3 Serializer

kylef:
Initial proof of concept with numerous caveats and unsupported elements (tracked in STATUS.md).

- [ ] Add smoke test for fury-cli

### DIFF
--- a/packages/openapi3-serializer/README.md
+++ b/packages/openapi3-serializer/README.md
@@ -1,0 +1,26 @@
+# API Elements: OpenAPI 3 Serializer
+
+[![NPM version](https://img.shields.io/npm/v/@apielements/openapi3-serializer.svg)](https://www.npmjs.org/package/@apielements/openapi3-serializer)
+[![License](https://img.shields.io/npm/l/@apielements/openapi3-serializer.svg)](https://www.npmjs.org/package/@apielements/openapi3-serializer)
+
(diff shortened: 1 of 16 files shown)


packages/openapi3-serializer/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Usage
+
+```js
+const fury = require('fury');
+const openapi3Serializer = require('@apielements/openapi3-serializer');
+
+fury.use(openapi3Serializer);
+
+// Assume `api` is a Minim element instance, e.g. from `fury.parse(...)`
+fury.serialize({ api, mediaType: 'application/vnd.oai.openapi' }, (error, content) => {
```
opichals, line 23: I'd expect the `serialize` would be synchronous for easy of use.

kylef: The `parse` and `serialize` functions in Fury are async. As we provide a uniform interface for serializing regardless of format and adapter implementation. This includes the ["remote" adapter](https://github.com/apiaryio/api-elements.js/tree/master/packages/fury-adapter-remote) which uses HTTP to parse/serialize this could be problematic to support them all as sync.

I don't have much strong opinions against or for this, so if you would like sync API we can create an issue in the repository to investigate support across all the adapters.

kylef: FWIW, master has sync interface for serialize now. We can support it here :smile:.

BenFradet:
Hey :wave: , might I ask what's the status of this PR? Thank you :+1:

kylef:
> Hey wave , might I ask what's the status of this PR? Thank you +1

This is more of a proof of concept, I'm not sure we'll pursue this any further.

PR #476, opened 2020-06-09
Cache body generation for allOf's with ref

kylef:
This continues on from https://github.com/apiaryio/api-elements.js/pull/409 but includes caching the case where a schema object uses `allOf` with single $ref. This appears to be a common case with some schema generators:

```yaml
allOf:
  - $ref: x
```

That is particullary because `$ref` cannot be used alongside other keys at root, the following can be invalid:

```yaml
$ref: '#/definitions/x'
definitions:
  x: {type: string}
```

Whereas this is acceptable:

```yaml
allOf: [{$ref: '#/definitions/x'}]
definitions:
  x: {type: string}
```

In documents utilising this particular case you can see up to 4x performance improvements:

```shell
# master
$ time npx fury a.json /dev/null
       54.86 real        41.67 user         1.98 sys

$ time npx fury b.json /dev/null
       23.94 real        20.30 user         0.92 sys

# this changeset

$ time npx fury a.json /dev/null
       13.34 real        12.79 user         0.80 sys

$ time npx fury b.json /dev/null
        7.12 real         7.00 user         0.44 sys
```

### DIFF
--- a/packages/openapi2-parser/lib/parser.js
+++ b/packages/openapi2-parser/lib/parser.js
@@ -1611,7 +1611,16 @@ class Parser {
     const referencedPathValue = this.referencedPathValue();
     let cacheKey;
     if (referencedPathValue && referencedPathValue.$ref) {
+      // schema object with $ref
       cacheKey = `${referencedPathValue.$ref};${contentType}`;
+    } else if (
+      referencedPathValue
+      && referencedPathValue.allOf
+      && Object.keys(referencedPathValue).length === 1
+      && referencedPathValue.allOf.length === 1
+      && referencedPathValue.allOf[0].$ref) {
+      // schema object with single ref in allOf (`allOf: [{$ref: path}]`)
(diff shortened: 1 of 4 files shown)


packages/openapi2-parser/lib/parser.js
```diff
@@ -1611,7 +1611,11 @@ class Parser {
     const referencedPathValue = this.referencedPathValue();
     let cacheKey;
     if (referencedPathValue && referencedPathValue.$ref) {
+      // schema object with $ref
       cacheKey = `${referencedPathValue.$ref};${contentType}`;
+    } else if (referencedPathValue && referencedPathValue.allOf && Object.keys(referencedPathValue).length === 1 && referencedPathValue.allOf.length === 1 && referencedPathValue.allOf[0].$ref) {
```
opichals, line 1616: I'd wrap the code somehow so that reviewing is easier (ideally no horizontal scrolling).

PR #479, opened 2020-06-10
feature: add support for generateMessageBody and schema

Mnau-mnau:


packages/remote/lib/adapter.js
```diff
@@ -73,17 +73,33 @@ class FuryRemoteAdapter {
...
       method: 'post',
       url: this.options.parseEndpoint,
       baseURL: this.options.url,
       data: source,
       params: {
         generateSourceMap,
+        adapterOptions: {
+          generateMessageBody: this.generateMessageBody,
+          generateMessageBodySchema: this.generateMessageBodySchema,
+        },
```
kylef, line 102: The options should be together as they should be in the querystring like the following:`?generateSourceMap=true&generateMessageBody=true`

```suggestion
        generateMessageBody: this.generateMessageBody,
        generateMessageBodySchema: this.generateMessageBodySchema,
```

packages/remote/lib/adapter.js
```diff
@@ -73,17 +73,31 @@ class FuryRemoteAdapter {
...
+    } else {
+      this.generateMessageBody = generateMessageBody;
+    }
+
+    if (generateMessageBodySchema === undefined) {
+      this.generateMessageBodySchema = true;
+    } else {
+      this.generateMessageBodySchema = generateMessageBodySchema;
+    }
+
```
kylef, line 91: I don' tthink this logic is needed, we can propergate it below directly without transformation and handle the defaults on the server side. If `generateMessageBody` is undefined, then we can prevent it being sent to the API as a parameter and let the API handle the default.

Below could become:

```
params: {
  generateSourceMap,
  generateMessageBody,
  generateMessageBodySchema,
}
```

At least I think that would work, perhaps I am mis understanding how axios is working.

PR #487, opened 2020-06-12
Validate OAS 2 documents contain swagger key

kylef:
In some cases, the document could be OpenAPI 3 and pass the swagger-parser validation. We should ensure the document is swagger 2.

### DIFF
--- a/packages/openapi2-parser/lib/parser.js
+++ b/packages/openapi2-parser/lib/parser.js
@@ -103,6 +103,15 @@ class Parser {
       return done(null, this.result);
     }
 
+    if (loaded.swagger === undefined) {
+      this.createAnnotation(
+        annotations.VALIDATION_ERROR, null,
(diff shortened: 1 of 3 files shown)


packages/openapi2-parser/lib/parser.js
```diff
@@ -103,6 +103,15 @@ class Parser {
       return done(null, this.result);
     }
 
+    if (loaded.swagger === undefined) {
+      this.createAnnotation(
+        annotations.VALIDATION_ERROR, null,
+        "Missing required key 'swagger'"
```
opichals, line 109: nit: should it be quoted with " or ' ? we use ' elsewhere.

also, above there is a string wrapped with parentheses... like `('xyz')`.

kylef: Agreed, using this style in openapi3-parser. Didn't realise the inconsistency at the time.

PR #491, opened 2020-06-19
Improve OpenAPI 2 validation error message

kylef:
When you have the following:

```yaml
type: foo
```

The following annotations are returned:

- error: (4) Data does not match any schemas from 'anyOf' - line 1
- error: (4) No enum match for: foo - line 1
- error: (4) Expected type array but found type string - line 1

Which is a bit confusing, we can do better, so this change makes it one error with a clearer message:

- error: (4) Value must be either 'array', 'boolean', 'integer', 'null', 'number', 'object', or 'string' not 'foo' - line 1

I realised this was causing user confusion in https://stackoverflow.com/questions/62468422/dredd-fails-to-run-with-errors-data-does-not-match-any-schemas-from-anyof-and .

I've also improved some other cases:

- when a schema validation fails with oneOf with two sub-validations. If that subvalidation is about missing required property `$ref` we will skip the missing ref property, the other validation error inside oneOf will be clearer in all cases I can think of.
- when a `oneOf` schema validation fails when all cases are about missing required properties in an object - we can collapse this all into one clearer message

Which results in the following improvements to the validation of a parameter object:

```yaml
in: query
name: foo
```

Where I am missing a required property of either `schema` or `type, which previous caused the following annotations previously:

- error: (4) Data does not match any schemas from 'oneOf' - line 8
- error: (4) Data does not match any schemas from 'oneOf' - line 8
- error: (4) Missing required property: schema
- error: (4) Missing required property: type
- error: (4) Missing required property: $ref - line 8

and now:

- error: (4) Object must contain either 'schema', or 'type' properties - line 8

This is a combination of refactoring, and fixes to the message. The commits are separated for clarity.


kylef:
This will need to be updated if https://github.com/apiaryio/api-elements.js/pull/492 is merged.

PR #495, opened 2020-06-19
Don't generate bodies when validating

kylef:
This brings a decent performance boost when you are validating OpenAPI 2 and do not want the parse result. The behaviour should remain unchanged, however we are disabling the generation of message bodies during validation as that is the slowest part of the parser.

Before:

```
$ time npx fury hbu/rsv.json --validate
        9.59 real        10.30 user         0.28 sys
```

After:

```
$ time npx fury hbu/rsv.json --validate
        2.83 real         3.23 user         0.11 sys
```

The added tests woth on both master and the branch, they are there to demonstrate the change retains prior behavior. When a Fury adapter does not implement validate, Fury will create one on top of parse. The caveat there is that it causes all the parse logic.

### DIFF
--- a/packages/openapi2-parser/lib/adapter.js
+++ b/packages/openapi2-parser/lib/adapter.js
@@ -29,9 +29,29 @@ function parse(options) {
   });
 }
 
+function validate(options) {
+  return new Promise((fulfil, reject) => {
+    const parser = new Parser({
+      ...options,
+      generateMessageBody: false,
+      generateMessageBodySchema: false,
+    });
+
+    parser.parse((error, parseResult) => {
+      if (error) {
+        reject(error);
+      } else if (parseResult.annotations.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/openapi2-parser/lib/adapter.js
```diff
@@ -29,9 +29,27 @@ function parse(options) {
   });
 }
 
+function validate(options) {
+  return new Promise((fulfil, reject) => {
+    const parser = new Parser(options);
+    parser.generateMessageBody = false;
+    parser.generateMessageBodySchema = false;
```
opichals, line 36: Seems strange to put those props directly on the parser. Why not the following?
```suggestion
    const parser = new Parser({
      ...options,
      generateMessageBody: false,
      generateMessageBodySchema: false
    });
```

PR #502, opened 2020-06-23
Prevent throwing during freeze for required annotations

kylef:
The parser can create multiple errors from an object which contain the exact same instance of a source map element. Thus creating integration problems using the parse result as the same element instance is re-used and thus a modification to one causes modification to others. This can be demonstrated by using `.freeze()` method as that requires each element to only be used a single time within the tree.

Prevents the error from being throw to users as per https://github.com/apiaryio/dredd/issues/1762

### DIFF
--- a/packages/openapi3-parser/test/unit/parser/parseObject-test.js
+++ b/packages/openapi3-parser/test/unit/parser/parseObject-test.js
@@ -18,6 +18,9 @@ describe('#parseObject', () => {
       name: 'Doe',
       message: 'Hello',
     });
+    object.attributes.set('sourceMap', [
+      new namespace.elements.SourceMap([[0, 10]]),
+    ]);
   });
 
   it('provides warning when given element is non-object', () => {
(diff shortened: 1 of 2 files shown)


packages/openapi3-parser/test/unit/parser/parseObject-test.js
```diff
@@ -141,6 +144,13 @@ describe('#parseObject', () => {
         "'Example Object' is missing required property 'required1'",
         "'Example Object' is missing required property 'required2'",
       ]);
+
+      // assert errors can be frozen (we're testing that the source map
+      // for object was cloned instead of referenced)
```
opichals, line 149: Out of curiosity how would it look like without the clone? It is not completely obvious for me how this is testing the not-a-reference.

kylef: `freeze()` would throw the following error:

> Cannot assign to read only property 'parent' of object '#<ArrayElement>'

kylef: That’s because effectively it’s doing this:

```
freeze() {
  freeze “this”
  
  for element in (all meta, attributes, content):
    element.parent = this
    element.freeze()
}
```

PR #503, opened 2020-06-23
Fix support for using `$ref` in a component definition

kylef:
I think it might be a bit hard for a reviewer to get their head around some of these changes as the referencing logic is some of the most complicated code in the project, so I'll try do my best at explaining these changes.

The following OpenAPI 3 document was not being parsed correctly:

```yaml
openapi: 3.0.0
info:
  title: example
  version: 1.0.0
paths: {}
components:
  schemas:
    User:
      type: object
    UserAlias:
      $ref: '#/components/schemas/User'
```

Support for using `$ref` directly in a reusable component against another reusable component wasn't supported. If users did that, they'd get incorrect behaviour and possibly incorrect errors (in the case for parameters which have required properties which a `$ref` object wouldn't contain).

To fix this I've made 3 changes which are broken into separate commits:

1. Before parsing, we create internal state of all the possible schemas. This allows us to parse schema A which references schema B (which isn't parsed yet), it's also part of supporting circular references. This is done by storing `context.state.components.set('schemas', ...)` where there is a collection of member elements with all the nessecery schema names. That way we can validate schemas exist. This particular code was specific to schemas only, in the first commit I have made this generic to apply to all reusable components. This internal state is needed to support references in other components.
2. Updated `parseReferenceObject` (which has dereferencing capabilities) to support multi-level referencing. This function supports returning "dereferenced" elements, this is used in areas such as `parseParametersObject` where parameter validation against a `href` uses the dereferenced parameter to get hold of the 'name'. Without this change, the function would return an element such as `{"element": "LimitAlias"}` instead the dereferenced limit parameter such as `{"element": "member", ...}`.
3. This is the core part of the fix, which is updating the logic that parses each component member to support referencing via the existing `parseReference` function. I've had to add the argument `returnReferenceElement` of `parseReferenceObject` to `parseReference` so it can be used here. Since we do actually want a referenced element back instead of a dereference element in this case (which isn't common with the rest of the uses of `parseReference`).

The last commit also contains integration tests to ensure this works end-to-end (for example JSON message body is created from the reference, etc). I want to ensure it works at all levels.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseComponentsObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseComponentsObject.js
@@ -8,6 +8,7 @@ const {
   createInvalidMemberWarning,
 } = require('../annotations');
 const parseObject = require('../parseObject');
+const parseReference = require('../parseReference');
 const pipeParseResult = require('../../pipeParseResult');
 const parseSchemaObject = require('./parseSchemaObject');
 const parseParameterObject = require('./parseParameterObject');
@@ -130,14 +131,17 @@ function parseComponentsObject(context, element) {
    * @returns ParseResult<ObjectElement>
    * @private
    */
-  const parseComponentObjectMember = (parser) => {
-    const parseMember = parseComponentMember(context, parser);
+  const parseComponentObjectMember = R.curry((parser, member) => {
+    const component = member.key.toValue();
 
-    return member => pipeParseResult(context.namespace,
+    const parseMember = parseComponentMember(context, parser);
+    const parseMemberOrRef = m => parseReference(component, () => parseMember(m), context, m.value, false, true);
+
+    return pipeParseResult(context.namespace,
       validateIsObject,
-      R.compose(parseObject(context, name, parseMember), getValue),
+      R.compose(parseObject(context, name, parseMemberOrRef), getValue),
       (object) => {
-        const contextMember = context.state.components.getMember(member.key.toValue());
+        const contextMember = context.state.components.getMember(component);
 
         if (contextMember) {
           contextMember.value = object;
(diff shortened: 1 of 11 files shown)


kylef:
This is related to https://github.com/apiaryio/dredd/issues/1762, the underlying document on this issue makes use of this feature of OpenAPI 3.0.

kylef:
I think there’s probably a case where this logic will get into an infinite loop, should be simple to resolve so adding this step

- [ ] Handle circular reference in the `dereference` function in `parseReferenceObject`

packages/openapi3-parser/lib/parser/oas/parseComponentsObject.js
```diff
@@ -130,12 +131,13 @@ function parseComponentsObject(context, element) {
    * @returns ParseResult<ObjectElement>
    * @private
    */
-  const parseComponentObjectMember = (parser) => {
+  const parseComponentObjectMember = R.curry((parser, member) => {
     const parseMember = parseComponentMember(context, parser);
+    const parseMemberOrRef = m => parseReference(member.key.toValue(), () => parseMember(m), context, m.value, false, true);
```
opichals, line 136: nit: I'd extract the `member.key.toValue()` to a variable to avoid calling it multiple times. It is also used below in this function.

packages/openapi3-parser/lib/parser/oas/parseComponentsObject.js
```diff
@@ -117,24 +131,23 @@ function parseComponentsObject(context, element) {
...
+    return pipeParseResult(context.namespace,
       validateIsObject,
-      R.compose(parseObject(context, name, parseMember), getValue),
+      R.compose(parseObject(context, name, parseMemberOrRef), getValue),
       (object) => {
         const contextMember = context.state.components.getMember(member.key.toValue());
 
         if (contextMember) {
           contextMember.value = object;
-        } else {
```
kylef, line 131: For review, this branch becomes dead code due to the first commit we only ever have the first case. This was evident in code coverage going down due to this line no longer being called/tested.

PR #506, opened 2020-06-25
Add accept header to requests

### LINKED ISSUE #413: OAS 3 parser does not generate Accept header's for the response content-type
Discovered via https://stackoverflow.com/questions/60264392/openapi3-and-csv-response-for-dredd.

```yaml
openapi: 3.0.0
info:
  title: example
  version: 1.0.0
paths:
  /:
    get:
      summary: csv
      responses:
        '200':
          description: test
          content:
            text/csv:
              schema:
                type: string
                example: 'cell1, cell2'

```

I'd expect the httpRequest element to include the Accept header, it does not:

```
$ npx fury issue.yaml | jq .content[0].content[0].content[0].content[0].content[0]
{
  "element": "httpRequest",
  "attributes": {
    "method": {
      "element": "string",
      "content": "GET"
    }
  }
}
```

kylef:
Discovered via https://stackoverflow.com/questions/60264392/openapi3-and-csv-response-for-dredd.

```yaml
openapi: 3.0.0
info:
  title: example
  version: 1.0.0
paths:
  /:
    get:
      summary: csv
      responses:
        '200':
          description: test
          content:
            text/csv:
              schema:
                type: string
                example: 'cell1, cell2'

```

I'd expect the httpRequest element to include the Accept header, it does not:

```
$ npx fury issue.yaml | jq .content[0].content[0].content[0].content[0].content[0]
{
  "element": "httpRequest",
  "attributes": {
    "method": {
      "element": "string",
      "content": "GET"
    }
  }
}
```

Fixes #413

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseOperationObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseOperationObject.js
@@ -47,6 +47,18 @@ function createTransactions(namespace, member, operation) {
       const clonedRequest = request.clone();
       clonedRequest.method = method;
 
+      const { contentType } = response;
+      if (contentType) {
+        let { headers } = clonedRequest;
+
+        if (!headers) {
+          headers = new namespace.elements.HttpHeaders();
+          clonedRequest.headers = headers;
+        }
+
+        headers.unshift(new namespace.elements.Member('Accept', contentType.toValue()));
+      }
+
       transactions.push(new namespace.elements.HttpTransaction([
         clonedRequest,
         response.clone(),
--- a/packages/openapi3-parser/CHANGELOG.md
+++ b/packages/openapi3-parser/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Adds support for the Schema Object title property.
 
+### Bug Fixes
+
+- Adds an accept header to requests with the matching responses content type.
+
 ## 0.14.0 (2020-06-24)
 
 ### Enhancements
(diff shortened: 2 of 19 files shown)


packages/openapi3-parser/lib/parser/oas/parseOperationObject.js
```diff
@@ -47,6 +47,18 @@ function createTransactions(namespace, member, operation) {
...
+      const { contentType } = response;
+      if (contentType) {
+        let { headers } = clonedRequest;
+
+        if (!headers) {
+          headers = new namespace.elements.HttpHeaders();
+          clonedRequest.headers = headers;
+        }
+
+        headers.unshift(new namespace.elements.Member('Accept', contentType.toValue()));
```
opichals, line 59: What if the response has an `Accept` header defined in the ADD?

kylef: `Accept` header is not permitted in Parameter Object (which is where else you could try set it) https://spec.openapis.org/oas/v3.0.2#fixed-fields-9

> If in is "header" and the name field is "Accept", "Content-Type" or "Authorization", the parameter definition SHALL be ignored.

We have test to ensure it's dropped and a warning is presented:

https://github.com/apiaryio/api-elements.js/blob/master/packages/openapi3-parser/test/unit/parser/oas/parseParameterObject-test.js#L136-L146

PR #514, opened 2020-07-20
Error out when parsing components using reserved elements

kylef:


packages/openapi3-parser/lib/parser/parseObject.js
```diff
@@ -20,6 +20,10 @@ const isAnnotationOrMember = R.anyPass([isAnnotation, isMember]);
  * @private
  */
 const chainParseResult = R.curry((transform, parseResult) => {
+  if (!parseResult.errors.isEmpty) {
```
opichals, line 23: Perhaps the `parseResultHasErrors` could be used here?

packages/openapi3-parser/lib/parser/parseObject.js
```diff
@@ -20,6 +20,10 @@ const isAnnotationOrMember = R.anyPass([isAnnotation, isMember]);
...
+    return parseResult;
+  }
+
   const result = R.chain(transform, parseResult);
 
   if (!result.errors.isEmpty) {
```
opichals, line 29: Also for `parseResultHasErrors` use?

PR #515, opened 2020-07-20
chore(oas2): recursive generateBody

opichals:
This makes the `generateBody` more likely to pick the `examples` defined in the ADD. e.g. `allOf[array[item]]`.

### DIFF
--- a/packages/openapi2-parser/lib/generator.js
+++ b/packages/openapi2-parser/lib/generator.js
@@ -26,7 +26,7 @@ function generateBody(schema) {
   }
 
   if (schemaIsArrayAndHasItems(schema)) {
-    return [generateBody(schema.items)];
+    return Array.from({ length: Math.min(5, schema.minItems || 1) }, () => generateBody(schema.items));
   }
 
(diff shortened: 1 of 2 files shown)


packages/openapi2-parser/lib/generator.js
```diff
@@ -17,18 +17,22 @@ faker.option({
   random: () => 0,
 });
 
-const schemaIsArrayAndHasItems = schema => schema.type && schema.type === 'array' && schema.items;
+const schemaIsArrayAndHasItems = schema => schema.type === 'array' && typeof schema.items === 'object';
```
kylef, line 20: `typeof schema.items === 'object'` is there a case where items wouldn't be an object? I think you'd get a parser error upfront when items is not an object:

> error: (4) Expected type object but found type array - line 31

opichals: The intent here was to make sure the `schemaIsArrayAndHasItems` return value is true/false. Also further down the recursion we rely on the fact (access `items.examples` property) that it is an object.

opichals:
@kylef amended the commit messages

PR #520, opened 2020-07-28
Prefer sample/default over empty primitive array when generating values

kylef:
In API Blueprint, the "value" of a type definition takes presidence over samples and defaults. For example take the following:

```apib
+ tags: one (array)
    + Sample
        + 1
        + 2
        + 3
```

-/-

We treat the followign JSON Schema (OAS Schema) roughly equivilent (as fixed-type):

```json
{
  "type": "array",
  "items": {
    "type": "string",
    "examples": ["one"]
  },
  "examples": [
    ["1", "2", "3"]
  ]
}
```

-/-

```json
{
  "element": "array",
  "attributes": {
    "samples": {
      "element": "array",
      "content": [
        {
          "element": "array",
          "content": [
            {
              "element": "string",
              "content": "1"
            },
            {
              "element": "string",
              "content": "2"
            },
            {
              "element": "string",
              "content": "3"
            }
          ]
        }
      ]
    }
  },
  "content": [
    {
      "element": "string",
      "content": "one"
    }
  ]
}
```

The above describes an array, with the value of an array containing `one` and an example `1`, `2`, and `3`. When retiving a value from an element, the value is preferred over samples and default. If a value is not present in the above array then it would fall back to samples and then default before generating an empty value.

This behaviour leads to a less ideal scenario where an empty value (such as primitive which does not contain its own value/samples/default) may produce an empty value disregarding samples or default. Such as with the following:

```apib
+ tags (array[string])
    + Sample
        + 1
        + 2
        + 3
```

-/-

We treat the followign JSON Schema (OAS Schema) roughly equivilent (as fixed-type):

```json
{
  "type": "array",
  "items": {
    "type": "string"
  },
  "examples": [
    ["1", "2", "3"]
  ]
}
```

Which manifests itself as a bug in https://github.com/apiaryio/dredd/issues/1796

-/-

```json
{
  "element": "array",
  "attributes": {
    "samples": {
      "element": "array",
      "content": [
        {
          "element": "array",
          "content": [
            {
              "element": "string",
              "content": "1"
            },
            {
              "element": "string",
              "content": "2"
            },
            {
              "element": "string",
              "content": "3"
            }
          ]
        }
      ]
    }
  },
  "content": [
    {
      "element": "string"
    }
  ]
}
```

The `valueOf` method would consider the array element is not empty, and then solely use the content elements to determine a value disregarding the samples. The change proposed here make it so that if the array contains primitive elements such as string, number or boolean which do not have a value of their own (or samples/default), then we will consider samples/defaults of the outter array before producing a value.

Before the above would result in `[""]` as a value because the value of a `StringElement` without value/samples/default/nullable ends up as an empty string `""` and thus is used to compute the value of the item in an array.

---

The first commit here is another problem I realised with the existing tests using incorrect samples definition. In API Elements the ArrayElement's content should contain an array, the existing code placed the sample value directly in array element's content which is incorrect form for an ArrayElement. It happens to be that these tests contain the sample values to ensure that the sample values are ignored and the content is not used thus they do affect test status.

### DIFF
--- a/packages/api-elements/lib/define-value-of.js
+++ b/packages/api-elements/lib/define-value-of.js
@@ -85,7 +85,16 @@ function findFirstSample(e) {
 
 const isPrimitive = e => (e instanceof StringElement) || (e instanceof NumberElement) || (e instanceof BooleanElement);
 const isEnumElement = e => e instanceof EnumElement;
-const isPlural = e => e instanceof ArrayElement;
+const isArrayElement = e => e instanceof ArrayElement;
+
+const hasSample = e => findFirstSample(e) !== null;
+const hasDefault = e => findDefault(e) !== null;
+const hasContent = e => e.content !== undefined;
+const hasValue = R.anyPass([hasContent, hasSample, hasDefault]);
+const hasNoValue = R.complement(hasValue);
+const isNoValuePrimitive = R.both(isPrimitive, hasNoValue);
+const isNonEmptyArrayElement = e => isArrayElement(e) && e.content && !e.isEmpty;
+
 
 function trivialValue(e) {
   if (e instanceof BooleanElement) {
@@ -113,8 +122,10 @@ function mapValue(e, options, f, elements) {
   }
 
   const opts = updateTypeAttributes(e, options);
-  if (e.content && (!isPlural(e) || !e.isEmpty)) {
+
+  if (e.content && !(isArrayElement(e) && e.content.every(isNoValuePrimitive))) {
     const result = f(e, opts, elements, 'content');
+
     if (undefined !== result) {
       return result;
     }
@@ -131,6 +142,15 @@ function mapValue(e, options, f, elements) {
   const dflt = findDefault(e);
   if (dflt) {
     const result = f(dflt, opts, elements, 'default');
+    if (undefined !== result) {
+      return result;
+    }
+  }
+
+  // reconsider content for array element (prefer sample/default first)
+  if (isNonEmptyArrayElement(e)) {
+    const result = f(e, opts, elements, 'content');
+
     if (undefined !== result) {
       return result;
     }
@@ -180,7 +200,7 @@ function mapValue(e, options, f, elements) {
     }
   }
 
-  if (isPlural(e) && e.isEmpty) {
+  if (isArrayElement(e) && e.isEmpty) {
     return f(e, opts, elements, 'generated');
   }
 
--- a/packages/api-elements/test/value-of-test.js
+++ b/packages/api-elements/test/value-of-test.js
@@ -700,6 +700,25 @@ describe('valueOf ArrayElement', () => {
     expect(value).to.deep.equal([4, 2]);
   });
 
+  it('prefers samples over empty primitive value', () => {
+    const element = new ArrayElement([new NumberElement()]);
+    element.attributes.set('default', new ArrayElement([4, 2]));
+    element.attributes.set('samples', new ArrayElement([
+      new ArrayElement([2, 'hello']),
+    ]));
+    const value = element.valueOf();
+
+    expect(value).to.deep.equal([2, 'hello']);
+  });
+
+  it('prefers default over empty primitive value', () => {
+    const element = new ArrayElement([new NumberElement()]);
+    element.attributes.set('default', new ArrayElement([4, 2]));
+    const value = element.valueOf();
+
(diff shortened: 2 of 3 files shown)


packages/api-elements/lib/define-value-of.js
```diff
@@ -87,6 +87,14 @@ const isPrimitive = e => (e instanceof StringElement) || (e instanceof NumberEle
 const isEnumElement = e => e instanceof EnumElement;
 const isPlural = e => e instanceof ArrayElement;
```
opichals, line 88: I'd name things based on what they do for clarity (perhaps in another PR)
```suggestion
const isArrayElement = e => e instanceof ArrayElement;
```

packages/api-elements/lib/define-value-of.js
```diff
@@ -87,6 +87,14 @@ const isPrimitive = e => (e instanceof StringElement) || (e instanceof NumberEle
...
 
+const hasSample = e => findFirstSample(e) !== null;
+const hasDefault = e => findDefault(e) !== null;
+const hasContent = e => e.content !== undefined;
+const hasValue = R.anyPass([hasContent, hasSample, hasDefault]);
+const hasNoValue = R.complement(hasValue);
+const isPrimitiveAndHasNoValue = R.both(isPrimitive, hasNoValue);
```
opichals, line 95:
```suggestion
const isNoValuePrimitive = R.both(isPrimitive, hasNoValue);
const isNonEmptyArrayElement = isArrayElement(e) && e.content && !e.isEmpty
```

packages/api-elements/lib/define-value-of.js
```diff
@@ -136,6 +146,15 @@ function mapValue(e, options, f, elements) {
     }
   }
 
+  // reconsider content for array element (prefer sample/default first)
+  if (isPlural(e) && e.content && !e.isEmpty) {
```
opichals, line 150:
```suggestion
  if (isNonEmptyArrayElement(e)) {
```

packages/api-elements/lib/define-value-of.js
```diff
@@ -113,8 +121,10 @@ function mapValue(e, options, f, elements) {
   }
 
   const opts = updateTypeAttributes(e, options);
-  if (e.content && (!isPlural(e) || !e.isEmpty)) {
+
+  if (e.content && !(isPlural(e) && (e.isEmpty || e.content.every(isPrimitiveAndHasNoValue)))) {
```
opichals, line 125: nit: I get this is does unnecessary e.content.every() call on empty content but I am trying to make all the conditions more readable

```suggestion
  if (e.content && !(isArrayElement(e) && e.content.every(isNoValuePrimitive))) {
```

packages/api-elements/test/value-of-test.js
```diff
@@ -24,9 +24,9 @@ describe('valueOf NullElement', () => {
   it('returns null with default and samples', () => {
     const element = new NullElement();
     element.attributes.set('default', new NullElement());
-    element.attributes.set('samples', new ArrayElement(
-      new NullElement()
-    ));
+    element.attributes.set('samples', new ArrayElement([
+      new NullElement(),
+    ]));
```
opichals, line 29: The `ArrayElement` constructor didn't change if I read the changes correctly. How do these test setup changes relate to the functionality change?

kylef: This was mentioned at the end of the PR description:

> The first commit here is another problem I realised with the existing tests using incorrect samples definition. In API Elements the ArrayElement's content should contain an array, the existing code placed the sample value directly in array element's content which is incorrect form for an ArrayElement. It happens to be that these tests contain the sample values to ensure that the sample values are ignored and the content is not used thus they do affect test status.

This is difference between:

```json
{
  “element”: “array”,
  “content”: { “element”: “string” }
}
```

and:

```json
{
  “element”: “array”,
  “content”: [
    { “element”: “string” }
  ]
}
```

opichals: > This was mentioned at the end of the PR description

🤦‍♂️ , sorry

packages/api-elements/lib/define-value-of.js
kylef: Fair

kylef:
@opichals I've applied all your suggestions, I'll squash this before merge but for now its separate commit to make review easier.

PR #526, opened 2020-07-31
Disallow non-string text body example in OpenAPI 3

kylef:
Discovered by @patricksmms that the included OpenAPI 3 document was emitting an invalid asset element. The asset element contained an object contents instead of a string as per API Elements spec. Asset element is a string element.

I've altered the parser behaviour so that it will warn user that the example is not a string and should be for text based media types. We will not produce a messageBody asset in this case, as we don't know how to serialise non-strings as text for text based media types. The OpenAPI 3 spec states the following for the example field:

> Example of the media type. The example object SHOULD be in the correct format as specified by the media type.

```yaml
openapi: 3.0.1
info:
  title: My API
  version: '2.0'
paths:
  /questions:
    post:
      summary: Description of the resource
      requestBody:
        content:
          application/json:
            schema:
              type: object
            example:
              message: hello world
      responses:
        '200':
          description: Created
          content:
            text/plain:
              schema:
                type: object
              example:
                message: hello world
```

`$ npx fury 526.yaml | jq .content[0].content[0].content[0].content[0].content[1].content[0]`:

```json
{
  "element": "asset",
  "meta": {
    "classes": {
      "element": "array",
      "content": [
        {
          "element": "string",
          "content": "messageBody"
        }
      ]
    }
  },
  "attributes": {
    "contentType": {
      "element": "string",
      "content": "text/plain"
    }
  },
  "content": [
    {
      "element": "member",
      "content": {
        "key": {
          "element": "string",
          "content": "message"
        },
        "value": {
          "element": "string",
          "content": "hello world"
        }
      }
    }
  ]
}
``

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseMediaTypeObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseMediaTypeObject.js
@@ -3,7 +3,7 @@ const contentTyper = require('content-type');
 const mediaTyper = require('media-typer');
 const pipeParseResult = require('../../pipeParseResult');
 const {
-  isExtension, hasKey, getKey, getValue,
+  isString, isExtension, hasKey, getKey, getValue,
 } = require('../../predicates');
 const {
   createWarning,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/openapi3-parser/lib/parser/oas/parseMediaTypeObject.js [hunk 1]
```diff
@@ -89,19 +89,28 @@ const createTextMessageBodyAsset = R.curry((namespace, mediaType, value) => {
...
+  const parseTextExample = pipeParseResult(namespace,
+    R.unless(isString, createExampleNotStringWarning),
+    createTextMessageBodyAsset(namespace, mediaType));
+
   return R.compose(R.cond([
     [() => isJSONMediaType(mediaType), createJSONMessageBodyAsset(namespace, mediaType)],
-    [() => isTextBasedType(mediaType), createTextMessageBodyAsset(namespace, mediaType)],
-    [R.T, createExampleNotJSONWarning],
+    [() => isTextBasedType(mediaType), parseTextExample],
+    [R.T, createExampleNotSupportedForMediaTypeWarning],
```
kylef, line 113: This old name doesn't reflect what the method does anymore, I think the name pre-dates adding support for text based media types.

PR #527, opened 2020-08-04
Marco/format link

marcofriso:
Closes APIARY-6105

### DIFF
--- a/packages/openapi2-parser/lib/parser.js
+++ b/packages/openapi2-parser/lib/parser.js
@@ -197,6 +197,7 @@ class Parser {
         // By default there are no groups, just the root API element
         this.group = this.api;
 
+        this.addFormatLink();
         this.handleSwaggerInfo();
         this.handleSwaggerHost();
         this.handleSwaggerAuth();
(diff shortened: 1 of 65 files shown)


packages/openapi2-parser/lib/parser.js
```diff
@@ -325,6 +326,18 @@ class Parser {
...
+  // Add the link to the format version specs
+  addFormatLink() {
+    const { Link } = this.namespace.elements;
+    const link = new Link();
+
+    link.title = 'OpenAPI 2';
+    link.relation = 'via';
+    link.href = 'https://spec.openapis.org/oas/v2.0';
+
+    return this.api.links.push(link);
```
kylef, line 338: The relation link should go into the parse result (`this.result`) directly instead of the API Category (`this.api`).

kylef: I didn't get a chance to elaborate on the explanation so just adding that now.

The reason behind it, is that the api category represents the API itself, whereas the parse result represents the result of parsing the source document. Thus sementically, the parse result came "via" the source format, but does not represent information about the API itself (only how it was described).

marcofriso:
I had similar errors locally today after I came back to this branch, I re-ran `env GENERATE=1 npm test` and now everything is passing again (locally)

kylef:
@marcofriso you need to commit all of the change fixtures, including the ones in `packages/swagger-zoo`.

marcofriso:
@kylef I re-installed everything with run and regenerated the fixtures, now I see the changes in `swagger-zoo`

PR #528, opened 2020-08-04
fix(oas3): attach description to hrefVariables' member

opichals:
Fix to make `hosts` category `resource` `hrefVariables` members conform to this https://github.com/apiaryio/api-elements.js/search?q=attaches+description+to+member&unscoped_q=attaches+description+to+member

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseServerVariableObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseServerVariableObject.js
@@ -50,7 +50,7 @@ const parseMember = context => R.cond([
  * @returns ParseResult<Member>
  * @private
  */
-const parseServerVariableObject = (context, element) => pipeParseResult(context.namespace,
+const parseServerVariableObject = (context, element, variableName) => pipeParseResult(context.namespace,
   R.unless(isObject, createWarning(context.namespace, `'${name}' is not an object`)),
   parseObject(context, name, parseMember(context), requiredKeys, [], true),
   (object) => {
@@ -58,11 +58,13 @@ const parseServerVariableObject = (context, element) => pipeParseResult(context.
 
     variable.attributes.set('default', object.get('default'));
 
+    const member = new context.namespace.elements.Member(variableName, variable);
+
     if (object.hasKey('description')) {
-      variable.description = object.getValue('description');
+      member.description = object.getValue('description');
     }
 
-    return variable;
+    return member;
   })(element);
 
 module.exports = R.curry(parseServerVariableObject);
--- a/packages/openapi3-parser/lib/parser/parseMap.js
+++ b/packages/openapi3-parser/lib/parser/parseMap.js
@@ -1,10 +1,12 @@
 const R = require('ramda');
 const pipeParseResult = require('../pipeParseResult');
 const {
-  isObject, getValue, isAnnotation,
+  isObject, getValue, isAnnotation, isMember,
 } = require('../predicates');
 const { createWarning } = require('./annotations');
 const parseObject = require('./parseObject');
+
+const isAnnotationOrMember = R.anyPass([isAnnotation, isMember]);
 
 const validateMapValueIsObject = (namespace, name, key) => R.unless(R.pipe(getValue, isObject), createWarning(namespace, `'${name}' '${key}' is not an object`));
 
@@ -15,8 +17,8 @@ const parseMapMember = R.curry((context, parser, member) => {
   const Member = R.constructN(2, context.namespace.elements.Member)(member.key);
 
   const parseResult = R.map(
-    R.unless(isAnnotation, Member),
-    parser(context, member.value)
+    R.unless(isAnnotationOrMember, Member),
+    parser(context, member.value, member.key)
   );
 
   if (isParseResultEmpty(parseResult)) {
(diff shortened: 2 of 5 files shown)


packages/openapi3-parser/lib/parser/oas/parseServerVariableObject.js
```diff
@@ -50,19 +50,21 @@ const parseMember = context => R.cond([
...
   (object) => {
     const variable = R.or(object.get('enum'), new context.namespace.elements.String());
 
     variable.attributes.set('default', object.get('default'));
 
+    const member = new context.namespace.elements.Member(variableName, variable);
+
     if (object.hasKey('description')) {
-      variable.description = object.getValue('description');
+      member.description = object.getValue('description');
```
kylef, line 64:
```suggestion
      member.description = object.get('description');
```

I think this should actually be `get` (should've been before too though), the difference here is `getValue` is roughly equivilent to `get(...).toValue()` where we would be putting the underlying string in the member instead of the element which would contain source maps.

This should "fix" missing source map for description of server variable objects.

packages/openapi3-parser/lib/parser/parseMap.js
```diff
@@ -1,11 +1,13 @@
 const R = require('ramda');
 const pipeParseResult = require('../pipeParseResult');
 const {
-  isObject, getValue, isAnnotation,
+  isObject, getValue, isAnnotation, isMember,
 } = require('../predicates');
 const { createWarning } = require('./annotations');
 const parseObject = require('./parseObject');
 
+const isAnnotationOrMember = R.anyPass([isAnnotation, isMember]);
```
kylef, line 9:
```suggestion
const isAnnotationOrMember = R.either(isAnnotation, isMember);
```

I'd prefer to use `either` over the more versatile `anyPass` as the implementation is simpler and more efficient. As we only have two items, it is clearer to use `either` semantically.

packages/openapi3-parser/test/unit/parser/oas/parseServerVariableObject-test.js
```diff
@@ -62,22 +63,22 @@ describe('#parseServerVariableObject', () => {
         description: 1234,
       });
 
-      const parseResult = parse(context)(serverVariable);
-      expect(parseResult.get(0)).to.be.instanceof(namespace.elements.String);
+      const parseResult = parse(context)(serverVariable, 'name');
+      expect(parseResult.get(1)).to.be.instanceof(namespace.elements.Annotation);
```
kylef, line 67: Fwiw, `expect(parseResult).to.contain.warning` (line below) checks that there is an annotation, and its class contains `warning`. The old line was testing that the first element is the valid element being returned (the precense of the warning still results in the first item being the expected value as apposed to a different warning or error). Generally all of the warning tests are written in same way to protect against the first value returning anything but a valid value.

packages/openapi3-parser/lib/parser/parseMap.js
```diff
@@ -15,8 +17,8 @@ const parseMapMember = R.curry((context, parser, member) => {
   const Member = R.constructN(2, context.namespace.elements.Member)(member.key);
 
   const parseResult = R.map(
-    R.unless(isAnnotation, Member),
-    parser(context, member.value)
+    R.unless(isAnnotationOrMember, Member),
```
kylef, line 20: I'm not going to block on it, however it might be a good idea to have coverage for this particular change in `parseMap-test`. I know it is tested through integration, although I've found that when changing (and breaking) these types of reusable components where behaviour is used upon many other places. It is a lot harder to figure out the case through debugging the integration.

packages/openapi3-parser/lib/parser/parseMap.js
opichals: This is a copy from elsewhere in the code :)

Can change all of them in another PR

kylef: Fair enough, yea I was using `anyPass` and `allPass` a lot before I discovered `either` and `both` 😄

packages/openapi3-parser/lib/parser/oas/parseServerVariableObject.js
opichals: Going to separate changing this into another PR

opichals:
merged the 'chore' commit

packages/openapi3-parser/lib/parser/oas/parseServerVariableObject.js
opichals: https://github.com/apiaryio/api-elements.js/pull/530

packages/openapi3-parser/lib/parser/parseMap.js
opichals: https://github.com/apiaryio/api-elements.js/pull/529

packages/openapi3-parser/lib/parser/parseMap.js
opichals: A test for this was added in https://github.com/apiaryio/api-elements.js/pull/528/commits/6f9ae7e49bf659a88a7ce3aa9998aabc2059f257

packages/openapi3-parser/test/unit/parser/oas/parseServerVariableObject-test.js
opichals: Kept the check for the valid entry in there.

PR #533, opened 2020-08-06
Marco/oas3 format link

marcofriso:
Closes APIARY-6106

### DIFF
--- a/packages/openapi3-parser/CHANGELOG.md
+++ b/packages/openapi3-parser/CHANGELOG.md
@@ -8,6 +8,10 @@
   This can dramatically increase the performance when dealing with documents
   which use many unsupported features.
 
+### Enhancements
+
+- added a Link element to the specific format/version in the parse result.
+
 ### Bug Fixes
 
 - Return a warning when parsing a document with a 'Media Type Object'
--- a/packages/openapi3-parser/lib/parser.js
+++ b/packages/openapi3-parser/lib/parser.js
@@ -126,7 +126,7 @@ function parse(source, context) {
     R.unless(isObjectOrAnnotation, createError(context.namespace, 'Source document is not an object')),
     R.unless(isAnnotation, parseOpenAPIObject(context)));
 
-  return R.chain(
+  const parseResult = R.chain(
     R.pipe(
       parseDocument,
       deduplicateUnsupportedAnnotations(context.namespace),
@@ -134,6 +134,22 @@ function parse(source, context) {
     ),
     document
   );
+
+  if (!isAnnotation(parseResult.content[0])) {
+    const formatVersion = (/3\.\d+\.\d+/).exec(source)[0];
+    const formatLink = `https://spec.openapis.org/oas/v${formatVersion}`;
+
+    const { Link } = context.namespace.elements;
+    const link = new Link();
+
+    link.title = `OpenAPI ${formatVersion}`;
+    link.relation = 'via';
+    link.href = formatLink;
+
+    parseResult.links.push(link);
+  }
+
+  return parseResult;
 }
 
 module.exports = parse;
--- a/packages/openapi3-parser/test/integration/fixtures/auth-scheme-does-not-exist.json
+++ b/packages/openapi3-parser/test/integration/fixtures/auth-scheme-does-not-exist.json
@@ -1,5 +1,31 @@
 {
   "element": "parseResult",
+  "meta": {
+    "links": {
+      "element": "array",
+      "content": [
+        {
+          "element": "link",
+          "meta": {
+            "title": {
+              "element": "string",
+              "content": "OpenAPI 3.0.0"
+            }
+          },
+          "attributes": {
+            "relation": {
+              "element": "string",
+              "content": "via"
+            },
+            "href": {
+              "element": "string",
+              "content": "https://spec.openapis.org/oas/v3.0.0"
+            }
+          }
+        }
+      ]
+    }
+  },
   "content": [
     {
       "element": "category",
--- a/packages/openapi3-parser/test/integration/fixtures/auth-scheme-does-not-exist.sourcemap.json
+++ b/packages/openapi3-parser/test/integration/fixtures/auth-scheme-does-not-exist.sourcemap.json
@@ -1,5 +1,31 @@
 {
   "element": "parseResult",
+  "meta": {
+    "links": {
+      "element": "array",
+      "content": [
+        {
+          "element": "link",
+          "meta": {
+            "title": {
+              "element": "string",
+              "content": "OpenAPI 3.0.0"
+            }
+          },
+          "attributes": {
+            "relation": {
+              "element": "string",
+              "content": "via"
+            },
+            "href": {
+              "element": "string",
+              "content": "https://spec.openapis.org/oas/v3.0.0"
+            }
+          }
+        }
+      ]
+    }
+  },
   "content": [
     {
       "element": "category",
--- a/packages/openapi3-parser/test/integration/fixtures/auth-scheme-global.json
+++ b/packages/openapi3-parser/test/integration/fixtures/auth-scheme-global.json
@@ -1,5 +1,31 @@
 {
   "element": "parseResult",
+  "meta": {
+    "links": {
+      "element": "array",
+      "content": [
+        {
+          "element": "link",
+          "meta": {
+            "title": {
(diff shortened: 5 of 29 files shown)


marcofriso:
Why are not there `package-lock.json` files? Should I add them to `.gitignore`?

kylef:
Yarn is used, not npm. There is yarn lockfile in root of repo, there is no need to run `npm install` here, you can delete the files and use yarn: (https://github.com/apiaryio/api-elements.js#using-lerna).

packages/openapi3-parser/lib/parser.js
```diff
@@ -126,14 +126,30 @@ function parse(source, context) {
...
     R.pipe(
       parseDocument,
       deduplicateUnsupportedAnnotations(context.namespace),
       context.options.generateSourceMap ? filterColumnLine : filterSourceMaps
     ),
     document
   );
+
+  if (!isAnnotation(parseResult.content[0])) {
+    const formatVersion = (/3\.\d+\.\d+/).exec(source)[0];
```
kylef, line 139: This can have false positives because it will run the regex on the entire source document, if the input was something along the lines of the following:

```yaml
info:
  title: hello
  version: 3.2.0
openapi: 3.0.0
paths: {}
```

This would incorrectly get `paths` as it is assuming that the first `3.x.x` value is the version of OpenAPI:

```
$ pbpaste | npx fury - | jq .meta.links.content[0]
{
  "element": "link",
  "meta": {
    "title": {
      "element": "string",
      "content": "OpenAPI 3.2.0"
    }
  },
  "attributes": {
    "relation": {
      "element": "string",
      "content": "via"
    },
    "href": {
      "element": "string",
      "content": "https://spec.openapis.org/oas/v3.2.0"
    }
  }
}
```

packages/openapi3-parser/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
   This can dramatically increase the performance when dealing with documents
   which use many unsupported features.
 
+### Enhancements
+
+- added a Link element to the specific format/version in the parse result.
+
```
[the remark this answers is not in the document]
marcofriso: yes

packages/openapi3-parser/lib/parser.js
marcofriso: inserted a "look-behind" into the regex

packages/openapi3-parser/lib/parser.js
```diff
@@ -126,14 +126,30 @@ function parse(source, context) {
...
     R.pipe(
       parseDocument,
       deduplicateUnsupportedAnnotations(context.namespace),
       context.options.generateSourceMap ? filterColumnLine : filterSourceMaps
     ),
     document
   );
+
+  if (!isAnnotation(parseResult.content[0])) {
+    const formatVersion = (/(?<=(["']?)openapi\1\s*:\s*(["']?))(3\.\d+\.\d+)/).exec(source)[0];
```
kylef, line 139: This regex still have false positives, I do not think it is a good idea to use regex for this and I would recommend follow my suggestion from before in the prior review (https://github.com/apiaryio/api-elements.js/pull/533#pullrequestreview-466602157) by using the parsed YAML/JSON output object instead of raw source. It will be more reliable.

This regex is not counting for all possible cases (note that the source can be either YAML or JSON). With some valid YAML inputs I am seeing this logic throw an error:

```yaml
"openapi": '3.0.0'
info:
  title: hello
  version: 3.2.0
paths: {}
```

```
TypeError: Cannot read property '0' of null
    at parse (/Users/kyle/Projects/apiaryio/api-elements.js/packages/openapi3-parser/lib/parser.js:139:92)
    at /Users/kyle/Projects/apiaryio/api-elements.js/packages/openapi3-parser/lib/adapter.js:29:21
    at new Promise (<anonymous>)
    at Object.parse (/Users/kyle/Projects/apiaryio/api-elements.js/packages/openapi3-parser/lib/adapter.js:25:10)
    at Fury.parse (/Users/kyle/Projects/apiaryio/api-elements.js/packages/core/lib/fury.js:230:29)
    at FuryCLI.run (/Users/kyle/Projects/apiaryio/api-elements.js/packages/cli/lib/fury.js:104:23)
    at Object.<anonymous> (/Users/kyle/Projects/apiaryio/api-elements.js/packages/cli/lib/fury.js:243:11)
    at Module._compile (internal/modules/cjs/loader.js:1158:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1178:10)
    at Module.load (internal/modules/cjs/loader.js:1002:32)
```

There is other cases for false positives:

Input:

```yaml
info:
  title: hello
  version: 3.2.0
paths:
  /parse:
    post:
      requestBody:
        content:
          application/openapi+json:
            example:
              openapi: 3.0.0
      responses:
        '204':
          description: result
openapi: 3.0.1
```

Incorrectly returns openapi 3.0.0 instead of 3.0.1:

```json
{
  "element": "link",
  "meta": {
    "title": {
      "element": "string",
      "content": "OpenAPI 3.0.0"
    }
  },
  "attributes": {
    "relation": {
      "element": "string",
      "content": "via"
    },
    "href": {
      "element": "string",
      "content": "https://spec.openapis.org/oas/v3.0.0"
    }
  }
}
```

marcofriso: ok, I misunderstood

marcofriso: changed https://github.com/apiaryio/api-elements.js/pull/533/commits/48abc84718745440b7d268d55ccb291a5467f16f

packages/openapi3-parser/lib/parser.js
```diff
@@ -126,14 +128,35 @@ function parse(source, context) {
...
     R.pipe(
       parseDocument,
       deduplicateUnsupportedAnnotations(context.namespace),
       context.options.generateSourceMap ? filterColumnLine : filterSourceMaps
     ),
     document
   );
+
+  if (!isAnnotation(parseResult.content[0])) {
+    const formatVersion = R.pipe(
```
kylef, line 141: I'd suggest adding a check here for the value of the openapi member element being a string element, and if any of these checks fail I think we should default to the latest OpenAPI 3 version we know about which is `3.0.3`. I think there may be cases where we could get here with an invalid document which cannot be parsed correctly.

Making a test similar to `it('add the format link', () => {` but instead the source is invalid document should verify this, if you set source to be an unexpected value such as `const source = '{}'` or `const source = '[]'` etc.

packages/openapi3-parser/lib/parser.js
```diff
@@ -126,14 +128,35 @@ function parse(source, context) {
...
+  const parseResult = R.chain(
     R.pipe(
       parseDocument,
       deduplicateUnsupportedAnnotations(context.namespace),
       context.options.generateSourceMap ? filterColumnLine : filterSourceMaps
     ),
     document
   );
+
+  if (!isAnnotation(parseResult.content[0])) {
```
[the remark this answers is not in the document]
marcofriso: This works a bit different than in API Blueprint version,
if I remove it it will cause 2 tests to fail.

```
  1) #parse
       fails to parse an OAS3 document with invalid YAML:
     TypeError: Cannot read property 'find' of undefined

  2) #parse
       fails to parse a non-object YAML document:
     TypeError: Cannot read property 'find' of undefined
```

This because in case of invalid Yaml there is no version of OAS3 to be parsed.
On the other hand, if there is a parse result containing errors, it will still be able to read the format version contained in the API Category (`parseResult.content[0]`) as Annotation will be (`parseResult.content[1]`).
How do you suggest to resolve this?

packages/openapi3-parser/lib/parser.js
marcofriso: changed here https://github.com/apiaryio/api-elements.js/pull/533/commits/f52cd031afdfb61aa2cf685c37dd758084f49d54

marcofriso: check if the value of the OpenAPi member element is a string element https://github.com/apiaryio/api-elements.js/pull/533/commits/88bcb5687ef5bf41922147ee51e1824da6c01830

packages/openapi3-parser/lib/parser.js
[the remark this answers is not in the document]
marcofriso: removed

PR #534, opened 2020-08-07
feat(apib): add format link

marcofriso:


packages/apib-parser/lib/adapter.js
```diff
@@ -42,9 +42,46 @@ function parse({
...
+      ],
+    },
+  };
+
+  return drafter.parse(source, options).then((result) => {
+    const refractElement = {};
+    refractElement.element = result.element;
+    refractElement.meta = formatLink;
+    refractElement.content = result.content;
+    return refractElement;
```
kylef, line 77: One caveat here is that if the underlying parser added any other meta or attributes they would go missing. It might be a bit simpler to construct a minim elements from the given result. In `options` there is `namespace` which can be used as a minim namespace. You can get hold of a ParseResult element instance in the same way that the other parsers work. Returning a ParseResult here does not change the behaviour, the API Elements Core logic detects if the returned value is a Minim element, in the cases it is not, it will convert it to be one (thus doing what you'd be doing here).

```js
const { namespace } = options;
const parseResult = namespace.fromRefract(source);
.. new namespace.elements.Link(...);
return parseResult;
```

What do you think?

marcofriso: @kylef in `options` there is not `namespace`,  I don't know how to access it here. Seems to me that API Elements are managed internally by the `drafter` function

kylef: The options passed into the function:

```

function parse({
  source, generateSourceMap, generateMessageBody, generateMessageBodySchema,
  requireBlueprintName,
}) {
```

can be adjusted to add `namespace`.

marcofriso: changed

packages/apib-parser/lib/adapter.js
```diff
@@ -42,9 +42,29 @@ function parse({
...
-  return drafter.parse(source, options);
+  return drafter.parse(source, options).then((result) => {
+    const parseResult = namespace.fromRefract(result);
+    const isAnnotation = element => element.element === 'annotation';
+    const { Link } = namespace.elements;
+
+    if (!isAnnotation(parseResult.content[0])) {
+      const link = new Link();
+
+      link.title = 'Apiary Blueprint';
```
kylef, line 53:
```suggestion
      link.title = 'API Blueprint';
```

This is the API Blueprint adapter

packages/apib-parser/lib/adapter.js
```diff
@@ -42,9 +42,29 @@ function parse({
     requireBlueprintName,
   };
 
-  return drafter.parse(source, options);
+  return drafter.parse(source, options).then((result) => {
+    const parseResult = namespace.fromRefract(result);
+    const isAnnotation = element => element.element === 'annotation';
+    const { Link } = namespace.elements;
+
+    if (!isAnnotation(parseResult.content[0])) {
```
kylef, line 50: Can you please explain the intent of this check? I am not sure it is needed.

marcofriso: In case of error it does not add the format link, do you want to remove it?

packages/apib-parser/lib/adapter.js
marcofriso: changed

packages/apib-parser/lib/adapter.js
kylef: I think in all cases we should include the link element, some editors such as Apiary show the source format on annotations which the link element's title will be useful for:

Screenshot 2020-08-17 at 17 03 48

The use-case may be finding the underlying format which caused the error. The "Unknown" here would become "OpenAPI" etc.

marcofriso: removed condition https://github.com/apiaryio/api-elements.js/pull/534/commits/e310ea38b517a25339e50e8b3c62679204790438

PR #537, opened 2020-08-12
chore: Moved to github ci

pksunkara:


pksunkara:
@kylef Do you have any idea about docs failure? I couldn't get pipenv running locally.

kylef:
> Do you have any idea about docs failure?

@pksunkara I think I've seen that exact problem before, and I think using an older Node (10) would solve it for the docs container. There was some compatibility issues with jsdoc (which upgrading breaks support for older Nodes too). We probably want to keep those node/pythons as close to what they are on readthedocs since that's where master doc builds will happen.

kylef:
https://github.com/apiaryio/api-elements.js/pull/374 was it

pksunkara:
@kylef Please review and merge.

PR #540, opened 2020-08-13
JSON serializer improvements

patricksmms:


packages/json-serializer/test/serializeJSON-test.js
```diff
@@ -24,9 +24,15 @@ describe('#serializeJSON', () => {
...
     new namespace.elements.Category([
       new namespace.elements.Category([
-        new namespace.elements.String('hello world', { id: 'message' }),
-      ], { classes: ['dataStructures'] }),
-      element,
+        new namespace.elements.Category([
+          new namespace.elements.Category([
+            element,
+          ], { classes: ['dataStructures'] }),
+        ], { classes: ['api-category'] }),
```
kylef, line 35:
```suggestion
        ], { classes: ['api'] }),
```

The class for an api category is just `api` (https://apielements.org/en/latest/element-definitions.html#id35)

packages/json-serializer/test/serializeJSON-test.js
```diff
@@ -22,16 +22,18 @@ describe('#serializeJSON', () => {
     const element = new namespace.elements.Element();
     element.element = 'message';
 
+    const dataStructure = new namespace.elements.Element();
+    dataStructure.element = 'dataStructure';
```
kylef, line 26:
```suggestion
    const dataStructure = new namespace.elements.DataStructure();
```

There’s a DataStructure element which can be used, that way line 26 won’t be needed (https://apielements.org/projects/js/en/latest/api.html#DataStructure).

packages/json-serializer/test/serializeJSON-test.js
```diff
@@ -22,11 +22,18 @@ describe('#serializeJSON', () => {
     const element = new namespace.elements.Element();
     element.element = 'message';
 
+    const dataStructure = new namespace.elements.DataStructure();
```
kylef, line 25: I'm not sure this tests the intended change because its still passing in `element` to `serializeJSON` instead of `dataStructure`. Perhaps it would be better to create a more specific test for it too like in https://github.com/apiaryio/api-elements.js/pull/544/commits/05c0c1e58d4d97ccc68d987f742a5142f8039e30

patricksmms: You are right. I added that scenario [here](https://github.com/apiaryio/api-elements.js/pull/540/commits/e570f3f8d4af57c6fcc090fc75e87e5b6fed2aab#diff-58e8df6f0d1b33d31ce6d546934715f8R41).

kylef (approved):

PR #546, opened 2020-08-20
feat(json): add indentation to serialize output

patricksmms:
I didn't add a parameter to set the indentation as, at least for now, there is no use case for it. Also it would require releasing the core.


kylef (approved):
@patricksmms Do you want to bump the version in package.json / CHANGELOG.md too? Should save a subsequent PR for releasing

patricksmms:
@kylef done

kylef (approved):

PR #549, opened 2020-08-27
feat(core): add serializeSync method

patricksmms:


packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # API Elements: Core
 
+## 0.2.1 (2020-08-27)
```
kylef, line 3: I think this is a breaking change to the contract between the core and the adapters and should be `0.3.0`

packages/core/lib/fury.js
```diff
@@ -244,6 +256,33 @@ class Fury {
...
+   * @param {Object} options
+   * @param {Category} options.api
+   * @param {string} [options.mediaType]
+   */
+  serializeSync({ api, mediaType = 'text/vnd.apiblueprint' }) {
+    const adapter = findAdapter(this.adapters, mediaType, 'serialize');
+
+    if (!adapter) {
+      // eslint-disable-next-line no-console
+      console.warn('Media type did not match any registered serializer!');
```
kylef, line 271: Should we make this throw so that it behaves the same way as the async counterpart?

```js
try {
    const result = await fury.serialize(...);
} catch (error) {
    //
}
```

```js
try {
    const result = fury.serializeSync(...);
} catch (error) {
    //
}
```

That way API consumers can catch the error and know the reason for failure programmatically.

packages/core/lib/fury.js
```diff
@@ -244,6 +256,33 @@ class Fury {
...
+      return null;
+    }
+
+    if (!api) {
+      // eslint-disable-next-line no-param-reassign
+      api = new this.minim.elements.Category();
+    }
+
+    return adapter.serialize({
+      api, namespace: this.minim, mediaType, sync: true,
```
kylef, line 282: Some adapters such as `apib-serializer` and `remote` are not respecting this `sync` flag and thus this API would return a promise.

kylef: Perhaps it might make sense to instead introduce a `serializeSync` method in the adapters. When calling sync version when the adapter doesn't implement that functionality, we could throw an error stating that the particular adapter doesn't support the action as synchronous.

Core could implement the "async" version to use `serializeSync` when `serialize` (async) isn't implemented by an adapter. In the similar way that the `validate` method can be fulfilled by `parse` above.

packages/core/CHANGELOG.md
patricksmms: The `serialize` method works well without passing the attribute `sync`. I'm adding it just to have it explicit.

packages/core/lib/fury.js
patricksmms: I thought about that. Just that if someone consumes it without wrapping in `try/catch` then it will terminate the control flow. It's not the equivalent of `Promise.reject`, also not something I would expect from this function.

In other hand the `console.warn` isn't nice either. I'm fine with changing as you suggest :)

packages/core/CHANGELOG.md
kylef: I'm referring more to this case described in https://github.com/apiaryio/api-elements.js/pull/549#discussion_r478458072, adapters designed for @apielements/core 0.2 won't respect `sync: true` (and would return promise). I think from these changes the expectation would be that if sync is set to true, a promise shouldn't be returned.

packages/apiaryb-parser/package.json
```diff
@@ -26,10 +26,10 @@
     "deckardcain": "^1.0.0"
   },
   "peerDependencies": {
-    "@apielements/core": ">=0.1.0 <0.3.0"
+    "@apielements/core": ">=0.1.0 <0.4.0"
```
kylef, line 29: As this adapters and others won't get release here, the current versions won't be compatible with the new core. I think we either need to release all the adapters, or make the change compatible with the old range.

Given the API has changed and now the `serialiseSync` is implemented in a way it will not break with old adapters (https://github.com/apiaryio/api-elements.js/pull/549#discussion_r478455762) then possibly going back to releasing core as `0.2.1` might make sense.

packages/core/lib/fury.js
patricksmms: Addressed in 41b4d5f

packages/core/lib/fury.js
patricksmms: Addressed in 66ab8f2 and fc81a6d

packages/apiaryb-parser/package.json
patricksmms: Reverted the version bump in 877bfb4, and added the only missing `serializeSync` in fc81a6d.

packages/apib-serializer/test/adapter-test.js
```diff
@@ -47,6 +47,30 @@ describe('API Blueprint serializer adapter', () => {
         return done();
       });
     });
+
+    it(`serializes synchronously ${path.basename(file)}`, (done) => {
```
kylef, line 51: Since this isn't an async API, perhaps it would be clearer to omit `done`. I think the test would be simpler without it since you won't need to wrap the blocks in try/catch as throwing from the test would trigger test failure.

packages/cli/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # API Elements: CLI Changelog
 
+## 0.10.3 (2019-08-31)
```
kylef, line 3: No strong opinions here, although we don't necessarily need to release the CLI as these changes have no impact on it. The new version of the apib serialize adapter is also compatible with the existing version `"@apielements/apib-serializer": "^0.16.1",` so installing CLI would get updated apib-serializer regardless.

kylef (approved):
Looks good to me, I've left some minor comments.

packages/apib-serializer/test/adapter-test.js
patricksmms: Absolutely

packages/cli/CHANGELOG.md
patricksmms: Right. I saw similar releases on the past, that's why I did it.

kylef: The past ones are usually a slightly different case where the version range didn't suffice (or core changed). Such as minor releases which don't fit into the patch range (0.17 for example won't fit into that range, and might trigger CLI release).

patricksmms: In that case I'll drop this release.

patricksmms: Dropped in e0d00c3

packages/apib-serializer/test/adapter-test.js
patricksmms: Addressed in cd072b1

packages/core/lib/fury.js
```diff
@@ -252,7 +287,7 @@ class Fury {
    */
 
   /**
-   * Serialize an API Description into the given output format.
+   * Serialize asynchronously an API Description into the given output format.
```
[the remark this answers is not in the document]
patricksmms: Addressed in 20ee86c

PR #555, opened 2020-09-07
feat(text): add text serializer

patricksmms:


packages/text-serializer/lib/adapter.js
```diff
@@ -0,0 +1,18 @@
+const serializeText = require('./serializeText');
+
+const name = 'text';
+const mediaTypes = [
+  'text/plain',
```
kylef, line 5: I think we should match on `text/*`, i.e, any string element containing `<html>...</html>` as content with the content-type `text/html` should work, etc.

This PR probably isn't the place to solve this, so I'd probably just put something in our backlog to solve in the future. I think this will require changing how the core works.

packages/text-serializer/test/adapter-test.js
```diff
@@ -0,0 +1,37 @@
...
+
+  it('has a name', () => {
+    expect(adapter.name).to.equal('text');
+  });
+
+  it('has a text/plain media type', () => {
+    expect(adapter.mediaTypes).to.deep.equal(['text/plain']);
+  });
+
+  it('can serialize an element asynchronously', (done) => {
```
kylef, line 21: I think it would be a good idea to have a test like this, but for the error case to count for the feedback above.

packages/text-serializer/lib/serializeText.js
```diff
@@ -0,0 +1,41 @@
...
+      if (isNotEmptyStringElement) {
+        dataStructures[element.id.toValue()] = element;
+      }
+    });
+  }
+
+  return dataStructures;
+}
+
+const isPrimitive = e => (e !== Object(e));
```
kylef, line 25: I think there might be one case where this won't work as intended, and that will be "undefined". Although I am not 100% sure if it is possible, but I think `valueOf` might return undefined in some cases where there isn't a value. Perhaps an enum element which doesn't contain any enumerations (`new namespace.elements.Enum()`) thus there isn't any possibly valid value to return.

Perhaps it would be good to have a test case for an empty enum to verify this case.

packages/text-serializer/lib/adapter.js
```diff
@@ -0,0 +1,18 @@
...
+];
+
+function serialize({ api }) {
+  return new Promise(resolve => resolve(serializeText(api)));
```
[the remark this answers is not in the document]
patricksmms: This:
```js
const syncFunc = (shouldTrow) => {
  if (shouldTrow) {
    throw new Error("oops");
  }

  return "success";
};

const asyncFunc = (shouldTrow) =>
  new Promise((resolve) => resolve(syncFunc(shouldTrow)));

asyncFunc(true)
  .then((result) => console.log("resolves with", result))
  .catch((err) => console.error("throws this", err));

asyncFunc(false)
  .then((result) => console.log("resolves with", result))
  .catch((err) => console.error("throws this", err));

```

Results into

```bash
resolves with success 
throws this Error: oops
```

packages/text-serializer/lib/serializeText.js
patricksmms: Addressed on 2a32fc3

packages/text-serializer/lib/adapter.js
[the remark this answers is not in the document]
patricksmms: It's a problem, though, if you don't use the `catch` method, as it would unexpectedly throw instead of rejecting the promise. I'll address it.

packages/text-serializer/test/adapter-test.js
patricksmms: Addressed in a131adf

packages/text-serializer/lib/adapter.js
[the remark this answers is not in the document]
patricksmms: Addressed in a131adf

packages/text-serializer/lib/serializeText.js
```diff
@@ -0,0 +1,42 @@
...
+
+const isPrimitive = element => element && element.element.match(/^(string|number|boolean|null)$/);
+
+function serializeText(element, done) {
+  if (element.element === 'dataStructure') {
+    return serializeText(element.content, done);
+  }
+
+  const dataStructures = collectElementByIDs(element);
+  if (isPrimitive(element) || isPrimitive(dataStructures[element.element])) {
```
kylef, line 33: Unfortunately I don't think checking that the element is primitive in this way would catch all cases. It might be better to go back to similar logic to what you had before where you are checking the result of valueOf. Since `valueOf` will handle all  other types of elements such as enum of string, ref element, or multiple levels of inheritance.

I think the following cases would demonstrate:

```js
const element = new Enum();
element.enumerations = ['north', 'east', 'south', 'west'];

element.valueOf(); // north (string)
```

Or more levels of recursion:

```js
const message = new StringElement('hello world');
message.id = 'Message';

const error = new Element();
error.element = 'message';
error.id = 'Error';

const notFoundError = new Element();
error.element = 'error';

notFoundError.valueOf(); // hello world (string)
// resolving notFoundError should resolve as a string from Message
```

There's possibly another case with `ref` element too.

patricksmms: Addressed in 2d3de2b

kylef (approved):

kylef (approved):

PR #556, opened 2020-09-07
fix(core): prefer sample/default over undefined props

patricksmms:
Prefer sample/default over Object with undefined property values

### DIFF
--- a/packages/api-elements/test/value-of-test.js
+++ b/packages/api-elements/test/value-of-test.js
@@ -968,6 +968,25 @@ describe('valueOf ObjectElement', () => {
     expect(value).to.deep.equal({ gaga: 'bing' });
   });
 
+  it('prefers samples over undefined property values', () => {
+    const element = new ObjectElement({ key1: new StringElement(), key2: new StringElement() });
+    element.attributes.set('default', new ObjectElement({ key1: 'defaultValue', key2: 'otherDefaultValue' }));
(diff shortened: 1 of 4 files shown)


packages/api-elements/test/value-of-test.js
```diff
@@ -968,6 +968,25 @@ describe('valueOf ObjectElement', () => {
...
+  });
+
+  it('prefers default over undefined property values', () => {
+    const element = new ObjectElement({ key1: new StringElement(), key2: new StringElement() });
+    element.attributes.set('default', new ObjectElement({ key1: 'defaultValue', key2: 'otherDefaultValue' }));
+    const value = element.valueOf();
+
+    expect(value).to.deep.equal({ key1: 'defaultValue', key2: 'otherDefaultValue' });
+  });
+
```
kylef, line 989: I'd suggest adding additional tests to cover the "value" being unset alongside the tests you have for value being set where `.content` is undefined. I.e:

```js
const element = new ObjectElement([
  new MemeberElement('key1'),
  new MemeberElement('key2'),
]);
```

`isObjectWithUndefinedValues` may need a check that `prop.value` isn't undefined before checking `prop.value.content`

patricksmms: Addressed in 8bb571e

kylef (approved):

PR #563, opened 2020-09-16
Move to GitHub Actions

kylef:
This completes moving CI to GitHub Actions. The documentation part was proposed in a separate PR before: https://github.com/apiaryio/api-elements.js/pull/562

The setup mostly works the same, however there is a slight difference to how coverage is setup. Previously we only uploaded coverage for Node 12, here I've made use of the functionality from Coveralls GitHub Action to report coverage for all the tests on each matrix node version. This reduces the amount of time for the coverage as we are not running a separate set of tests specific for coverage as each test run gets coverage. I made an attempt at trying to get coverage for the smoke tests too, but given how it uses `npm pack` and a different set of sources it proved cumbersome so I gave up on that part.


.github/workflows/test.yaml
```diff
@@ -0,0 +1,42 @@
+on:
+  push:
+    paths-ignore:
+      - 'docs/**'
+      - 'README.md'
```
kylef, line 5: Biggest reason to have "lint" in a separate file is that we can skip running unit/smoke tests for documentation only changes, but the commit linting still applies.

patricksmms:
Are you leaving the circeci configs intentionality?

kylef:
@patricksmms I'm not, they should be removed (https://github.com/apiaryio/api-elements.js/pull/563/commits/67380707f43b2398cf9b7c0ea82d120e162641ce) / (https://github.com/apiaryio/api-elements.js/tree/kylef/gh-actions)

patricksmms:
> @patricksmms I'm not keeping them intentionally, they should be removed (https://github.com/apiaryio/api-elements.js/pull/563/commits/67380707f43b2398cf9b7c0ea82d120e162641ce) / (https://github.com/apiaryio/api-elements.js/tree/kylef/gh-actions)

Sorry, on the mobile app it's hardly visible when checking the changes 😄

patricksmms (approved):
LGTM

patricksmms (approved):

PR #564, opened 2020-09-21
fix(core): fix structure type attributes

patricksmms:
This PR is to sort out the attributes functionalities.

This was implemented before, but was broken https://github.com/apiaryio/api-elements.js/pull/561/files#diff-ed3f4907fe76fdf2e90227edd1bba42cL70

### DIFF
--- a/packages/api-elements/lib/define-value-of.js
+++ b/packages/api-elements/lib/define-value-of.js
@@ -26,34 +26,22 @@ const {
 } = require('./utils');
 
 /**
- * @typedef InheritedAttrs
- * @type {object}
- * @property {boolean} isFixed
- * @property {boolean} isFixedType
- * @property {boolean} isNullable
- */
-
-/**
  * Map the element values
  * @param {element} e - element
- * @param {InheritedAttrs} inheritedAttrs - inherited attributes
+ * @param {boolean} inheritFixed - inherited fixed attribute
  * @param {function} f - map function
  * @param {object=} elements - object map of elements to look for inherited types
  * @return {any}
  */
-function mapValue(e, f, inheritedAttrs, elements) {
+function mapValue(e, f, inheritFixed, elements) {
   if (e === undefined) {
     return undefined;
   }
 
-  const attrs = {
-    isFixed: inheritedAttrs.isFixed || isFixed(e),
-    isFixedType: (inheritedAttrs.isFixed && isPrimitive(e)) || isFixedType(e),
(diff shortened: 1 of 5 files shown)


packages/api-elements/lib/define-value-of.js
```diff
@@ -24,20 +25,35 @@ const {
...
  */
-function mapValue(e, f, elements) {
+function mapValue(e, f, inheritedAttrs, elements) {
   if (e === undefined) {
     return undefined;
   }
 
+  const attrs = {
+    isFixed: inheritedAttrs.isFixed || isFixed(e),
+    isFixedType: (inheritedAttrs.isFixed && isPrimitive(e)) || isFixedType(e),
```
kylef, line 51: Fixed-type isn't an inherited type attribute, I think this part should be removed?

kylef (changes requested):

packages/api-elements/lib/define-value-of.js
patricksmms: Indeed. Addressed in ef0df9c.

I'm now realizing it's missing tests for [required](https://github.com/apiaryio/api-elements.js/pull/564/files#diff-ed3f4907fe76fdf2e90227edd1bba42cR158).

kylef (approved):

packages/api-elements/lib/utils.js
```diff
@@ -32,11 +32,18 @@ function hasTypeAttribute(e, attribute) {
 }
 
 /**
- * Check if element has 'fixed' or 'fixedType' typeAttribute set
+ * Check if element has 'fixed' typeAttribute set
  * @param {element} e - element
  * @return {boolean}
  */
-const isFixed = e => hasTypeAttribute(e, 'fixed') || hasTypeAttribute(e, 'fixedType');
+const isFixed = e => hasTypeAttribute(e, 'fixed');
```
kylef, line 39: Unrelated to PR but just something I noticed while reading this again, would be nice to make `hasTypeAttribute` curried and in opposite argument order. I can make a PR once these are merged

```
const isFixed = hasTypeAttribute('fixed');
```

kylef (approved):
Looks good to me!

packages/api-elements/lib/utils.js
patricksmms: I can do that on the [other PR](https://github.com/apiaryio/api-elements.js/tree/patricksmms/improve-define-value-of)

PR #567, opened 2020-10-12
Initial OpenAPI 3.1 support

kylef:
These changes include initial support for some OpenAPI 3.1 changes, there's a lot more, this change is more to explore the way the parser will operate and conditionally support 3.1 features.

The following changes have been done:

> responses are no longer required to be defined under the Operation Object.
> Introduced a new top-level field - webhooks. This allows describing out-of-band webhooks that are available as part of the API.

We don't fully support `webhooks`, but with 3.1 the appropriate warning will be shown.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseOpenAPIObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseOpenAPIObject.js
@@ -38,6 +38,8 @@ function parseOASObject(context, object) {
     return new namespace.elements.ParseResult([array].concat(parseResult.annotations.elements));
   };
 
+  const isOpenAPI31OrHigher = () => context.isOpenAPIVersionMoreThanOrEqual(3, 1);
+
   const parseMember = R.cond([
     [hasKey('openapi'), parseOpenAPI(context)],
     [hasKey('servers'), R.compose(parseServersArray(context, name), getValue)],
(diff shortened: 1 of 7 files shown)


packages/openapi3-parser/lib/parser/oas/parseOpenAPIObject.js
```diff
@@ -49,14 +51,18 @@ function parseOASObject(context, object) {
...
+    ],
     [isUnsupportedKey, createUnsupportedMemberWarning(namespace, name)],
 
     // Return a warning for additional properties
     [R.T, createInvalidMemberWarning(namespace, name)],
   ]);
 
   const parseOASObject = pipeParseResult(namespace,
-    parseObject(context, name, parseMember, requiredKeys, ['components']),
+    parseObject(context, name, parseMember, requiredKeys, ['openapi', 'components']),
```
kylef, line 65: To aid review, the array of keys passed to `parseObject` here signify an order of keys that should be parsed upfront (in order). This means that `openapi` is the first key to be parsed as other code will depend on it being parsed. Thus for documents such as the following:

```
webhooks: {}
openapi: 3.1.0
```

The parse order will actually be `openapi` then `webhooks`.

opichals (approved):
Looks good. 🚀

PR #585, opened 2021-02-11
Simple support for Parameter Object's schema

kylef:
This change brings very primitive support for the schema field in a Parameter Object. Full support is not that simple because there are numerous limitations in both API Elements and the surrounding tooling (Apiary v3 documentation, Apiary Style Guides) which may not expect references and more complex structures found in parameter value's. Typically parameter values are a flat structure.

Thus this is a very simple support for the Parameter Object schema field, only the `type` and `example` are supported only. `$ref` or any other property will produce warnings.

There is a little bit of duplication between the existing schema parser and the parameter object specific one, I don't think it is worth refactoring the old schema object to be more generic for this one-off case, and no-doubt if there was future changes to it, it would make it harder to achieve without having to consider the implications in parameters. The long term goal would be removing parameter specific schema parser to use a single schema parser. This can be a good first step.

These changes took me 10 mins in my free time to produce this, if there are major feedback against then I would just close this instead of investing any further time into it (imo its not worth it). By all means, that doesn't mean I don't want to put this in without good review :smile:.

With propagation to Dredd it would solve https://github.com/apiaryio/dredd/issues/1363.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseParameterObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseParameterObject.js
@@ -12,12 +12,13 @@ const pipeParseResult = require('../../pipeParseResult');
 const parseObject = require('../parseObject');
 const parseString = require('../parseString');
 const parseBoolean = require('../parseBoolean');
+const parseSchemaObject = require('./parseParameterObjectSchemaObject');
 
 const name = 'Parameter Object';
 const requiredKeys = ['name', 'in'];
 const unsupportedKeys = [
   'deprecated', 'allowEmptyValue', 'style', 'allowReserved',
-  'schema', 'examples', 'content',
+  'examples', 'content',
 ];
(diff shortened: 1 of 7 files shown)


opichals (approved):
Looks great! Thanks! :rocket:

packages/openapi3-parser/lib/parser/oas/parseParameterObject.js
```diff
@@ -166,6 +168,14 @@ function parseParameterObject(context, object) {
...
 
+      if (example === undefined) {
+        const schema = parameter.get('schema');
+
+        if (schema) {
+          member.value = schema;
+        }
+      }
+
       const description = parameter.get('description');
```
opichals, line 179: @kylef Is the lack of setting `member.title` intentional in OAS parsers?

The `drafter` APIB parser seems to be [setting 'title' to the parameter type](https://github.com/apiaryio/drafter/blob/master/packages/drafter/src/RefractAPI.cc#L209)

PR #587, opened 2021-02-11
Fix missing source map for media type examples >1 warning

### LINKED ISSUE #519: OpenAPI 3 media type examples warning misses source map
```
warning: 'Media Type Object' 'examples' only one example is supported, other examples have been ignored
```

kylef:
Fixes #519

### DIFF
--- a/packages/openapi3-parser/test/integration/fixtures/regression/519-examples-warning-source-map.json
+++ b/packages/openapi3-parser/test/integration/fixtures/regression/519-examples-warning-source-map.json
@@ -1,0 +1,203 @@
+{
+  "element": "parseResult",
+  "content": [
+    {
+      "element": "category",
+      "meta": {
+        "classes": {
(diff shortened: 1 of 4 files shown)


packages/openapi3-parser/test/integration/fixtures/regression/519-examples-warning-source-map.json
```diff
@@ -0,0 +1,203 @@
...
+          "content": [
+            {
+              "element": "string",
+              "content": "warning"
+            }
+          ]
+        }
+      },
+      "attributes": {
+        "sourceMap": {
```
kylef, line 156: This entire block would have been missing because the code previously used the member element as the source of the warning and member elements do not contain source maps from the YAML parser. We are not using the key of the member (`two` from the example).

opichals (approved):
LGTM!

PR #592, opened 2021-02-18
Support parameter object schema title/description

kylef:


packages/openapi3-parser/lib/parser/oas/parseParameterObjectSchemaObject.js
```diff
@@ -91,6 +91,9 @@ function parseSchemaObject(context) {
     [hasKey('example'), e => e.clone()],
     [hasKey('enum'), R.compose(parseEnum(context, name), getValue)],
 
+    [hasKey('title'), parseString(context, name, false)],
```
kylef, line 94: The third argument is to indicate if the string is optional or required. Title is not required thus `false`, the existing test ensure this works correctly because they do not have title/description.

opichals (approved):
:+1:

PR #594, opened 2021-02-19
Validate server variable enum is not empty

kylef:
This was from OpenAPI 3.1 rc1 changelog:

> Server Variable's enum now MUST not be empty (changed from SHOULD).

Technically before it was also a should and we didn't warn. The behaviour in our parser is that MUST's generally behave as warn and discard the value instead of erroring out completely (to be less hostile). So the behaviour is same for both versions in this parser.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseServerVariableObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseServerVariableObject.js
@@ -20,9 +20,15 @@ const parseEnumStrings = R.curry((context, element) => {
   const parseEnumString = R.unless(isString,
     createWarning(namespace, `'${name}' 'enum' array value is not a string`));
 
+  const validateArrayNotEmpty = R.when(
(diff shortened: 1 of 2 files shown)


packages/openapi3-parser/lib/parser/oas/parseServerVariableObject.js
```diff
@@ -20,9 +20,15 @@ const parseEnumStrings = R.curry((context, element) => {
   const parseEnumString = R.unless(isString,
     createWarning(namespace, `'${name}' 'enum' array value is not a string`));
 
+  const validateArrayNotEmpty = R.when(
+    array => array.isEmpty,
```
kylef, line 24: As this may be suggested in review, `R.isEmpty` won't work because it doesn't appear to work for ArrayElement.

opichals (approved):
LGTM :rocket:

PR #595, opened 2021-02-19
Warn for pathItems in components on 3.1

kylef:
3.1.0-rc0 brings:

> Components Object now has a new entry pathItems, to allow for reusable Path Item Objects to be defined within a valid OpenAPI document

I looked into supporting this but it will require a bit of refactoring to do first because right now the pathItemsParser returns concrete values which use the member's key (the path itself). Whereas we won't know that upfront when parsing reusable path items as they don't have a "path" (key) until use.

Path Item Object parser already has warning for `$ref` as use.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseComponentsObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseComponentsObject.js
@@ -224,6 +224,7 @@ function parseComponentsObject(context, element) {
       return parseResult;
     });
 
+  const isOpenAPI31OrHigher = () => context.isOpenAPIVersionMoreThanOrEqual(3, 1);
   const parseMember = R.cond([
     [hasKey('schemas'), parseSchemas],
     [hasKey('parameters'), parseComponentObjectMember(parseParameterObject)],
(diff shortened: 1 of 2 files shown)


kylef:
> I looked into supporting this but it will require a bit of refactoring to do first because right now the pathItemsParser returns concrete values which use the member's key (the path itself). Whereas we won't know that upfront when parsing reusable path items as they don't have a "path" (key) until use.

https://github.com/apiaryio/api-elements.js/pull/127 contains a similar refactoring when I did the same for responses object. The bigger difference with this one is that full parameter validation requires the path to validate the path variables are present in the path. That will have to be moved to the integration point instead of being self contained within the path item parser.

I.e, because the following:

```yaml
paths:
  /users/{id}:
    $ref: '#/components/pathItems/ListPathItem'
  /teams/{id}:
    $ref: '#/components/pathItems/ListPathItem'

components:
  pathItems:
    ListPathItem:
      parameters:
        in: path
        name: test
        required: true
```

The path parameter names can only be validated upon their integration with a path in `paths` section.

opichals (approved):
LGTM

PR #596, opened 2021-02-19
Warn for unsupported 3.1 security scheme mutalTLS

kylef:


kylef:
In terms of future support:

- Verify that Apiary v3 (metamorphoses docs don't explode with unknown/unsupported security schemes)
- Design the API Element structure for mTLS
- Parse it

opichals (approved):
LGTM

PR #598, opened 2021-02-22
Normalise OpenAPI type as an array internally

kylef:
This is one of the first steps in supporting OpenAPI 3.1's type as an array. This first step normalises the type into an array internally, so that all the internal code paths in the code base use arrays.

This means that the `parseType` function in the schema parser will always return an array of strings. In this implementation that array will always be of a single item. There is code which would throw if that wasn't the case. It is not possible to enter the code path, this will be altered in the future to something along the lines of:

```diff
@@ -271,7 +270,7 @@ function parseSchema(context) {
       } else if (enumerations) {
         element = enumerations;
       } else if (type.length > 1) {
-        throw new Error('Implementation error: unexpected multiple types');
+        element = type.map(type => constructStructure(namespace, schema, type));
       } else if (type.length === 1) {
         element = constructStructure(namespace, schema, type[0]);
       } else {
```

but first, the parser needs to be able to parse the arrays, and there will be various tests for all those cases. That will come later.

One thing I will point out, is that `parseString` accepts a member element and not a string, that's because it uses the member element's key to construct the error message, this makes it hard to compose with other subparsers such as `parseArray` which is why the string parsing is done inline. `parseType` accepts a string element (and in the future an array of string elements if you use OpenAPI 3.1). In the past, `parseType` accepted the member element with the value of a string element.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
@@ -6,7 +6,7 @@ const {
 } = require('../annotations');
 const pipeParseResult = require('../../pipeParseResult');
 const {
-  isString, hasKey, hasValue, getValue,
+  isString, hasKey, getValue,
 } = require('../../predicates');
 const parseObject = require('../parseObject');
 const parseArray = require('../parseArray');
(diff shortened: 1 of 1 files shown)


packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -94,16 +118,14 @@ function parseType(context) {
...
-      getValue
-    )
+    element => isValidType(element.toValue()),
+    createWarning(context.namespace, `'${name}' 'type' must be either ${types.join(', ')}`)
   );
 
   return pipeParseResult(context.namespace,
-    parseString(context, name, false),
-    ensureValidType);
+    R.unless(isString, value => createWarning(context.namespace, `'${name}' 'type' is not a string`, value)),
```
kylef, line 126: Eventually this branch will head into a direction where we handle multiple types, I'd imagine it will use `R.cond` such as:

```js
R.cond([
  [isString, ensureValidType, then cocerce to array],
  [isArray, ...],
  [R.T, createWarning(value must be ...)],
])
```

opichals (approved):
:ok_hand:

PR #599, opened 2021-02-22
Include nullable with Schema Object without a type on OAS 3.1

kylef:
In OpenAPI 3.1, `null` is permitted in `type`. This causes the behaviour when you have a schema without a type restriction to allow null. This wasn't the case in OpenAPI 3.0 or older, because these version removes `null` from JSON Schema's type system.


opichals (approved):
:+1:

PR #600, opened 2021-02-22
Support array of type in Schema Object

kylef:
The caveat is that the array can only be a single value. This brings the logic to parse and validate an array of types, but only allows one at the moment so it does not add additional functionality to the rest of the parser. This allows reviewing the array parsing, and all the validation cases independently of support for multiple types.

I'd perhaps suggest looking at the test cases first to understand all the desired behaviours.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
@@ -6,7 +6,7 @@ const {
 } = require('../annotations');
 const pipeParseResult = require('../../pipeParseResult');
 const {
-  isString, hasKey, getValue,
+  isArray, isString, hasKey, getValue,
 } = require('../../predicates');
 const parseObject = require('../parseObject');
 const parseArray = require('../parseArray');
@@ -128,10 +128,14 @@ const typeToElementNameMap = {
  * Normalises result into an ArrayElement of StringElement
  */
 function parseType(context) {
+  const { namespace } = context;
+
   let types;
   let isValidType;
 
-  if (context.isOpenAPIVersionMoreThanOrEqual(3, 1)) {
+  const isOpenAPI31OrHigher = R.always(context.isOpenAPIVersionMoreThanOrEqual(3, 1));
+
+  if (isOpenAPI31OrHigher()) {
     types = openapi31Types;
     isValidType = isValidOpenAPI31Type;
   } else {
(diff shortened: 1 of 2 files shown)


packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -121,13 +125,71 @@ function parseType(context) {
...
+  );
+
+  const parseArrayTypeItem = pipeParseResult(namespace,
+    R.unless(isString, createWarning(namespace, `'${name}' 'type' array value is not a string`)),
+    ensureValidTypeInArray);
+
+  const isEmpty = arrayElement => arrayElement.isEmpty;
+
+  // ArrayElement -> ParseResult<ArrayElement, Annotation>
+  const ensureTypesAreUnique = (types) => {
```
kylef, line 147: I know this is a bit complicated, I am pretty sure I can make this function a bit simpler tomorrow with fresh eyes.

packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -121,13 +125,71 @@ function parseType(context) {
...
+    return parseResult;
+  };
+
+  const parseArrayType = pipeParseResult(namespace,
+    R.when(isEmpty, createWarning(namespace, `'${name}' 'type' array must contain at least one type`)),
+    parseArray(context, `${name}' 'type`, parseArrayTypeItem),
+    ensureTypesAreUnique,
+
+    // FIXME support >1 type
+    R.when(e => e.length > 1, createWarning(namespace, `'${name}' 'type' more than one type is current unsupported`)));
```
kylef, line 175: These two lines will be removed to allow multiple types with the type parser in the future.

packages/openapi3-parser/test/unit/parser/oas/parseSchemaObject-test.js
```diff
@@ -60,114 +70,206 @@ describe('Schema Object', () => {
...
+        });
+        const parseResult = parse(context, schema);
+
+        expect(parseResult).to.contain.warning(
+          "'Schema Object' 'type' array must contain unique items, string is already present"
+        );
+      });
+
+      it('warns when array items contain more than 1 entry', () => {
+        // FIXME support more than one entry :)
```
kylef, line 249: This is the one test that is more temporary as it should go away once we do support multiple types.

packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -108,10 +108,14 @@ const typeToElementNameMap = {
  * Normalises result into an ArrayElement of StringElement
  */
 function parseType(context) {
+  const { namespace } = context;
+
   let types;
   let isValidType;
 
-  if (context.isOpenAPIVersionMoreThanOrEqual(3, 1)) {
+  const isOpenAPI31OrHigher = R.always(context.isOpenAPIVersionMoreThanOrEqual(3, 1));
```
opichals, line 116: Perhaps a suggestion for another PR

A `context.isOpenAPI31OrHigher` could be added as that is something we are using the `isOpenAPIVersionMoreThanOrEqual` almost always for IIRC. Then it could be destructured nicely in
```js
const { namespace, isOpenAPI31OrHigher } = context;
```

packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
opichals: What about using `R.uniqBy`?

opichals:
```js
const types = ['string', 'boolean', 'boolean', 'annotation'].map(t => ({ toValue: () => t }));

const toValue = R.invoker(0, 'toValue');
R.pipe(
  R.uniqBy(toValue),
  R.map(toValue)
)(types)
```

opichals: Obviously the above doesn't help collecting the Annotations separately. Perhaps you could get that single list along with annotations and then filter them out in the last step:

```js
const validated = R.map(validateType, types);
const annotation = R.find(is(Annotation), validated);
const validatedTypes = R.filter(complement(equals(annotation)), validated);
return new namespace.elements.ParseResult(validatedTypes, annotation);
```

packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -121,13 +125,71 @@ function parseType(context) {
...
+    }, types);
+
+    const parseResult = new namespace.elements.ParseResult();
+    parseResult.push(permittedTypes.elements);
+
+    if (warnings.length > 0) {
+      parseResult.push(...warnings);
+    }
+    return parseResult;
+  };
```
opichals, line 167: After the attempts I suggested above I think this way the implementation shows what you meant by the comment clearly so that I'd leave as is...

opichals (approved):

PR #601, opened 2021-02-22
Support OpenAPI 3.1 Schema Object const

kylef:
This is similar to enum, for a single value. This is equivalent to MSON / API Elements "fixed" types.


opichals (approved):
:rocket:

PR #605, opened 2021-02-23
Support `null` with other types on OpenAPI 3.1

kylef:
* Removes nullable on OAS 3.1 (this is breaking change in 3.1 spec)
* Supports type with array of 1 item alongside null (but no other case -- that will come)

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
@@ -192,7 +192,10 @@ function parseType(context) {
     ensureTypesAreUnique,
 
     // FIXME support >1 type
-    R.when(e => e.length > 1, createWarning(namespace, `'${name}' 'type' more than one type is current unsupported`)));
+    R.unless(
+      e => e.length === 0 || e.length === 1 || (e.length === 2 && e.contains('null')),
+      createWarning(namespace, `'${name}' 'type' more than one type is current unsupported`)
+    ));
 
   return R.cond([
     [isString, parseStringType],
@@ -213,10 +216,10 @@ function parseType(context) {
 }
 
 // Returns whether the given element value matches the provided schema type
-const valueMatchesType = (type, value) => {
+const valueMatchesType = R.curry((value, type) => {
   const expectedElementType = typeToElementNameMap[type];
   return value.element === expectedElementType;
-};
+});
 
 // Returns whether the given element value matches an enumeration of fixed values
 const valueMatchesEnumerationValues = (enumeration, value) => {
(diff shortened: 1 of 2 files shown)


packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -404,8 +412,9 @@ function parseSchema(context) {
         element.description = description;
       }
 
+      // On OAS 3.0, nullable is a keyword, on OAS 3.1, null goes in type
       const nullable = schema.getValue('nullable');
-      if (nullable) {
+      if (nullable || (type.includes('null') && element.element !== 'null')) {
```
kylef, line 417: We want to add nullable, only if the type isn't null.

`type: null` causes null element, whereas `type: [string, null]` means string element, that this branch makes nullable.

packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
```diff
@@ -192,7 +192,7 @@ function parseType(context) {
     ensureTypesAreUnique,
 
     // FIXME support >1 type
-    R.when(e => e.length > 1, createWarning(namespace, `'${name}' 'type' more than one type is current unsupported`)));
+    R.when(e => e.length > 1 && !e.contains('null'), createWarning(namespace, `'${name}' 'type' more than one type is current unsupported`)));
```
opichals, line 195: ['string', 'boolean', 'null'] would fail this

kylef: Very good point, I've applied this update:

```diff
diff --git a/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js b/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
index 1e3ac22..7dd7595 100644
--- a/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
@@ -192,7 +192,10 @@ function parseType(context) {
     ensureTypesAreUnique,
 
     // FIXME support >1 type
-    R.when(e => e.length > 1 && !e.contains('null'), createWarning(namespace, `'${name}' 'type' more than one type is current unsupported`)));
+    R.unless(
+      e => e.length === 0 || e.length === 1 || e.length == 2 && e.contains('null'),
+      createWarning(namespace, `'${name}' 'type' more than one type is current unsupported`)
+    ));
 
   return R.cond([
     [isString, parseStringType],
```

Normally I'd make a test for that case, but.. I am preparing a next PR that makes it supported so unless there's push back I'll not add a negative test for that case.

opichals (approved):
Thanks!

PR #597, opened 2021-02-22
Support null 3.1 type in Schema Object

kylef:

PR #606, opened 2021-02-23
Support multiple types in Schema Object

kylef:
This completes full support for multiple types in OpenAPI 3.1 schemas, this builds up on the work from other recent changes:

- [Support null 3.1 type in Schema Object ](https://github.com/apiaryio/api-elements.js/pull/597)
- [Normalise OpenAPI type as an array internally](https://github.com/apiaryio/api-elements.js/pull/598) - this brings some upfront refactoring so the code base deals with array of types. The internal interfaces are updated, with assumption that type can never be more than one.
- [Support array of type in Schema Object](https://github.com/apiaryio/api-elements.js/pull/600) - brings array of types to our public interface (i.e, support from input), but restricts the types to only allow 1. That changes brings all the logic for parsing arrays, duplicate detection, and the various test cases for parsing the array.
- [Support `null` with other types on OpenAPI 3.1](https://github.com/apiaryio/api-elements.js/pull/605) - brings support for specifying a type alongside null (and removes nullable for OpenAPI 3.1 which is replaced by null). Null is more of a special case in terms of types because for most part, in API Elements null alongside another type is that type with a nullable attribute.

### DIFF
--- a/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
+++ b/packages/openapi3-parser/lib/parser/oas/parseSchemaObject.js
@@ -86,7 +86,7 @@ function constructArrayStructure(namespace, schema) {
   return element;
 }
 
-function constructStructure(namespace, schema, type) {
+const constructStructure = R.curry((namespace, schema, type) => {
   let element;
 
   if (type === 'object') {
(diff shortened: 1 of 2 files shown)


opichals (approved):
:heart: :rocket:

PR #610, opened 2021-02-23
Support boolean additionalProperties in OpenAPI 2

kylef:
Boolean additional properties, for example (`additionalProperties: false` which indicates no additional properties are permittied), in API Elements that is an object with the `fixed-type` typeAttribute.

`additionalProperties` may also contain a subschema which all undeclared properties must conform to, I'll try prepare a follow up with that. That can become tricky due to the implications with the `required` keyword.


opichals (approved):
:+1:

PR #611, opened 2021-02-23
Support additionalProperties with subschema

kylef:
I'd suggest reviewing the tests first, they show the interactions with required which may be clearer when reading tests than the code.

### DIFF
--- a/packages/openapi2-parser/lib/schema.js
+++ b/packages/openapi2-parser/lib/schema.js
@@ -114,16 +114,29 @@ class DataStructureGenerator {
     // Create member elements for required keys which are not defined in properties
     const missingRequiredProperties = required.filter(name => properties[name] === undefined);
     const requiredMembers = missingRequiredProperties.map((name) => {
-      const member = new this.minim.elements.Member(name);
+      let member;
+
+      if (schema.additionalProperties !== undefined && typeof schema.additionalProperties !== 'boolean') {
+        member = this.generateMember(name, schema.additionalProperties);
+      } else {
+        member = new this.minim.elements.Member(name);
+      }
+
       member.attributes.set('typeAttributes', ['required']);
       return member;
     });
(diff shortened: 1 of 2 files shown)


packages/openapi2-parser/lib/schema.js
```diff
@@ -114,16 +114,29 @@ class DataStructureGenerator {
...
       member.attributes.set('typeAttributes', ['required']);
       return member;
     });
     element.content.push(...requiredMembers);
 
-    if (schema.additionalProperties === false) {
+    if (schema.additionalProperties !== undefined && schema.additionalProperties !== true) {
       const typeAttributes = element.attributes.get('typeAttributes') || new this.minim.elements.Array([]);
-      typeAttributes.push('fixed-type');
+      typeAttributes.push('fixedType');
```
kylef, line 132: I mistakenly got the case incorrect, `fixed-type` is the human readable name in MSON, `fixedType` is the casing for API Elements.

packages/openapi2-parser/test/schema-test.js
```diff
@@ -720,7 +720,90 @@ describe('JSON Schema to Data Structure', () => {
...
+      };
+
+      const dataStructure = schemaToDataStructure(schema);
+
+      expect(dataStructure.element).to.equal('dataStructure');
+      expect(dataStructure.content).to.be.instanceof(ObjectElement);
+
+      expect(dataStructure.content.length).to.equal(1);
+
+      const member = dataStructure.content.getMember('');
```
kylef, line 763: Typically, the key of the member is a "sample" key, but we don't really have one so I have used an empty string.

Note this structure is equivalent to API Blueprint, except the key name where API Blueprint provides a sample `name`.

```apib
+ Response 200 (application/json)
    + Attributes (fixed-type)
        + *name* (string)
```

opichals (approved):
:+1:

PR #641, opened 2021-07-15
Prevent treating overlapping paths as a circular reference 

kylef:
A check that a reference to `#/definitions/User` from within `#/definitions/UserList` would incorrectly been interpreted as a circular reference as a simplistic check that the current path starts with the current reference. We should ensure that the reference is either identical, or relative from another (such as with a trailing slash).

This was easiest to test at a unit level so I have added unit tests for this method. The top 4 added tests would already pass, just increasing test coverage to account for all the branches in the circular reference checker.

To aid review, the circular reference checker is passed all the references we've traversed (this is the first argument). If we would follow a series of references (such as if we are currently at `#/items`, and then we followed a reference to `#/definitions/User` which inside `#/definitions/User/properties/comments/items` referenced ` `#/definitions/Comment`. The current path would become `#/definitions/Comment` and the current traversed path would be something like `['#/items', '#/definitions/User', '#/definitions/User/properties/comments/items']`. These parts of the current resolution are pushed and popped as we resolve a schema.

### DIFF
--- a/packages/openapi2-parser/CHANGELOG.md
+++ b/packages/openapi2-parser/CHANGELOG.md
@@ -1,5 +1,14 @@
 # API Elements: OpenAPI 2 Parser Changelog
 
+## 0.32.6 (2021-07-21)
+
+### Bug Fixes
+
+- Fixes generating example JSON bodies from Swagger Schema when the schema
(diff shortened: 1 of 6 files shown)


opichals (approved):
Thanks, LGTM

PR #654, opened 2023-05-17
Update ramda to 0.29

kylef:
### What:

* Update ramda to 0.29 in api-elements core and the openapi3-parser.
* Updates CI config so that GitHub actions pass
* Install pegjs by https instead of git protocol clone. Git protocol is fairly old (modern HTTP protocol and SSH based protocols are typically used). In particular GitHub has changed the removed support for cloning by git protocol (https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git):
  > On the Git protocol side, unencrypted git:// offers no integrity or authentication, making it subject to tampering. We expect very few people are still using this protocol, especially given that you can’t push (it’s read-only on GitHub). We’ll be disabling support for this protocol.

#### Details

In ramda 0.28 a breaking change to to the `R.filter` and `R.reject` functions had been introduced, in the past `filter` would call `filter` on the underlying object, whereas now it will run fantasy land `filter` (as of https://github.com/ramda/ramda/pull/3002).

What's fantasy land and why does this matter?

Fantasy land is somewhat a specification (or set of laws) that a set of values should behave as, for example it states that `filter` should return the same type as the object that it was called upon. I.e, if you call `filter` on a `ParseResult`, it must return `ParseResult`.

The why it matters part, is that minim's "Array" and "ParseResult" structures implement both `filter` and fantasy land compatible filter. It does so, with a difference in behaviour https://github.com/refractproject/minim/commit/6274d467ecb0698b60b04cca2c92b29f88acae9c. `filter` returns an ArraySlice, whereas the fantasy land filter must return the same type.

What's the difference?

- An ArraySlice is a small structure which places elements in the slice under the `elements` property.
- An Element (the base class of Array and ParseResult and others) places all children elements under `content`.

### Why:

Older versions of ramda are flagged as insecure.

### Criticality:

Medium - Due to security nature.

### Blast radius:

Medium - The OpenAPI 3 parser is not type safe, although it has fairly good test coverage. Ramda a fundamental dependency, and changes to its behaviour can have widespread impact. Although I am fairly confident there are no widespread bugs introduced given the test coverage is very high (wrote most of the parser test-first/TDD).

### Expected outcome:

The version of ramda used is no longer flagged as insecure.

### Potential impact and visibility:

No expected impact.

### Test results:

The openapi3-parser has fairly wide test coverage that gives me enough confidence this change works as expected, at-least within the context of this project. Consumers of the library will need to uptake any ramda behaviour changes also.


.github/workflows/docs.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
           node-version: 10
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.6
+          python-version: 3.8
```
kylef, line 13: 3.6 is no longer available on GitHub actions. 3.8 isn't the latest either but there are breaking changes which would mean we have to uptake other changes that I wanted to do right now.

.github/workflows/test.yaml
```diff
@@ -31,12 +31,3 @@ jobs:
       - uses: actions/checkout@v2
       - run: yarn install --frozen-lockfile --ignore-optional
       - run: ./scripts/smoke.sh
-
-  coverage:
-    needs: [test]
-    runs-on: ubuntu-latest
-    steps:
-      - uses: coverallsapp/github-action@master
```
kylef, line 39: The GitHub org doesn't let us use external actions outside our space. Dropping coveralls reports (nice to have, and not fundamental).

packages/openapi3-parser/lib/parser.js
```diff
@@ -104,7 +104,7 @@ const deduplicateUnsupportedAnnotations = R.curry((namespace, parseResult) => {
     },
     R.T);
 
-  const result = new namespace.elements.ParseResult(R.filter(filterWarnings, parseResult));
+  const result = R.filter(filterWarnings, parseResult);
```
kylef, line 107: As fantasy land filter would result in the same type, we no longer need to wrap it back into a parse result.

packages/openapi3-parser/lib/parser/oas/parseOauthFlowsObject.js
```diff
@@ -72,7 +72,8 @@ function parseOauthFlowsObject(context, object) {
       authScheme.push(new namespace.elements.Member('grantType', grantTypes[member.key.toValue()]));
       authScheme.push(member.value.getMember('scopes'));
 
-      R.filter(R.is(namespace.elements.Transition), member.value).forEach((item) => {
+      const isTransitionElement = R.is(namespace.elements.Transition);
+      member.value.filter(isTransitionElement).forEach((item) => {
```
kylef, line 76: This one was a little tricky to grapple with. `member.value` here is an `ObjectElement`. An object element is a series of key/value pairs called members.

ObjectElement's [forEach has the signature of `(value, key, item)`](https://github.com/refractproject/minim/blob/master/lib/primitives/ObjectElement.js#L197). Thus when `filter` returns an ObjectElement, the behaviour changed. I have switched to `member.value.filter` which is equivalent to the prior behaviour as it would call `forEach` on ArraySlice which iterates over each member element and doesn't have the forEach semantics of an object.

